PR #28, opened 2014-10-31
Fix StackOverflowError in CassandraAdminTemplate#dropTable(Class<?>)

mp911de:
Deletes recursive function to determine table name from entity class and reuse getTableName(...)

Reference: DATACASS-174


prowave:
@matthewadams  - Can you take a look at this one since this is your area.  Thanks.

mp911de:
Merged with 71a1ba5.

PR #29, opened 2015-01-27
get on varint Column should return BigInteger not Integer

petercable:


petercable:
https://jira.spring.io/browse/DATACASS-199

PR #33, opened 2015-04-13
Remove redundant dependency on Liquibase from spring-cql

wilkinsona:
Commit 54910a1 added a dependency on org.liquibase:liquibase-core however it does not appear to be used anywhere and mvn clean package completes successfully without it.

The presence of this dependency causes problems in Spring Boot as it activates auto-configuration of Liquibase when a user does not expect it. See http://stackoverflow.com/questions/29594497/spring-boot-spring-boot-starter-data-jpa-database-type-none-or-cannot-find-cha, for example.

### DIFF
--- a/spring-cql/pom.xml
+++ b/spring-cql/pom.xml
@@ -114,11 +114,6 @@
 			<artifactId>guava</artifactId>
 			<version>15.0</version>
 		</dependency>
-		<dependency>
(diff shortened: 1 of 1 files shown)


olivergierke:
That's applied and merged back to Fowler and Evans bugfix branches.

PR #34, opened 2015-05-06
DATACASS-182 | fixed column values null'ing

walec51:


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/CassandraRepositoryFactory.java
```diff
@@ -70,7 +71,7 @@ public CassandraRepositoryFactory(CassandraOperations cassandraOperations) {
 
 	@Override
 	@SuppressWarnings({ "rawtypes", "unchecked" })
-	protected Object getTargetRepository(RepositoryMetadata metadata) {
+	protected Object getTargetRepository(RepositoryInformation metadata) {
```
walec51: you had an syntax error here in master, had to fix to run all tests

thomasdarimont: Hi, this was a result of simplified repository customizations - I created the more narrow focussed PR #35 to fix this.

Cheers,
Thomas

walec51:
this solution did not work, we got 
com.datastax.driver.core.exceptions.InvalidQueryException: Multiple definitions found for column
in tests

PR #39, opened 2015-06-15
Update CQLOperationsTest.java

kalmanh:
Move book creation after the check if the row is null


mp911de:
Thanks for submitting a pull request. The proposed changes do not add features, fix a bug or provide a benefit for Spring Data Cassandra so I'm closing this PR. If you are interested in contributing bugfixes or features, have a look in our [issue tracker](https://jira.spring.io/browse/DATACASS-281?jql=project%20%3D%20DATACASS%20AND%20resolution%20is%20EMPTY).

kalmanh:
The issue here is pretty simple (although minor). If `r` is `null` that means there is no data and the method returns. In this case, there is no reason to create a `Book` instance.

PR #43, opened 2015-11-24
DATACASS-239 - Latest cassandra-unit and dse-driver

prowave:
Updated Cassandra-Unit, DSE Driver and Cassandra version to latest
stable version.  Changed the Test Cases to support the API changes on
the driver by DataStax.  They changed the exception hierarchy for failed
replication factor quorum.  The embedded spring-cassandra.yaml files
were tuned to turn of unneeded features that were causing overhead in
the testing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -191,12 +191,14 @@
 				<groupId>org.apache.commons</groupId>
 				<artifactId>commons-lang3</artifactId>
 				<version>3.1</version>
+				<scope>test</scope>
 			</dependency>
 
 			<dependency>
 				<groupId>org.hectorclient</groupId>
 				<artifactId>hector-core</artifactId>
 				<version>1.1-4</version>
+				<scope>test</scope>
 			</dependency>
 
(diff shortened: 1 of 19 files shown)


pom.xml
```diff
@@ -190,6 +186,19 @@
 				<version>${jodatime}</version>
 				<scope>test</scope>
 			</dependency>
+
+			<dependency>
+				<groupId>org.apache.commons</groupId>
```
olivergierke: Do we really need these new dependencies? I am especially worried about the Commons Lang version as it's very likely to be an application dependency as well so that version conflicts are likely. We should try to avoid additional dependencies if possible.

prowave: Good catch.  I added the test scope to the dependencies that I had to add to get the tests working as a result of changes in the upstream dependencies.  Does that suffice?  The deps are 100% required for the tests to compile and run.

olivergierke:
That's merged, thanks Dave!

PR #44, opened 2015-12-21
queryForObject() Entity Mapping

ahmetkarakaya:
What I have found is that following method returns value of first column of row cassandraOps.queryForObject(s, Person.class) has the following method trace.

//only prints the first column of row
System.out.println("Person->"+cassandraOps.queryForObject(s, Person.class)); 

CqlTemplate-> processOne()

@SuppressWarnings("unchecked")
@Override
public <T> T processOne(ResultSet resultSet, Class<T> requiredType) throws DataAccessException {
    if (resultSet == null) {
        return null;
    }
    Row row = resultSet.one();
    if (row == null) {
        return null;
    }
    return (T) firstColumnToObject(row);
}


mp911de:
Hi @ahmetkarakaya,
this PR looks broken, at least I can't see any commits from you. Can you either fix the PR or close it?

mp911de:
Closing this PR as of no activity and branch 1.3.x won't get merged to master.

PR #45, opened 2015-12-23
fix java driver document link broken

boliza:
fix java driver document link broken


mp911de:
Thanks, that's merged with 2b503ec.

PR #47, opened 2016-03-18
DATACASS-260 - Support mapping for enum-typed properties.

mp911de:
We now support out-of-the box mapping for enum-typed properties in domain types and primary keys. Enums are mapped using their name. Ordinal-based enum mapping is not supported but can be achieved by registering a custom converter. Enums default to `DataType.Name.VARCHAR`. Type overrides can be achieved by using `@CassandraType` and supplying an own `Converter` to map `enum`s according to their `ordinal()` value.

```java
@Table("book")
public class Book {
 
	@PrimaryKey
	private String isbn;
	
	private Condition condition;
 
	// getter/setter omitted.
}
 
public enum Condition {
	MINT, MHHKAY, NOT_SO_GOOD;
}
```

----
Ticket: [DATACASS-260](https://jira.spring.io/browse/DATACASS-260)

### DIFF
--- a/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/integration/convert/MappingCassandraConverterTests.java
+++ b/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/integration/convert/MappingCassandraConverterTests.java
@@ -1,0 +1,361 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/integration/convert/MappingCassandraConverterTests.java
```diff
@@ -0,0 +1,361 @@
...
+import com.datastax.driver.core.querybuilder.Insert;
+import com.datastax.driver.core.querybuilder.QueryBuilder;
+import com.datastax.driver.core.querybuilder.Update;
+import com.datastax.driver.core.querybuilder.Update.Assignments;
+
+/**
+ * Tests for {@link MappingCassandraConverter}.
+ * 
+ * @author Mark Paluch
+ * @soundtrack Outlandich - Dont Leave Me Feat Cyt (Sun Kidz Electrocore Mix)
```
jxblum: What's this?

jxblum:
@mp911de I think these code changes and tests look good.

Ship It!

mp911de:
Merged via 5e09ef5adf62a9fbf6493be3b044fc2719dda554

PR #41, opened 2015-08-13
DATACASS-231 Update CassandraCqlClusterFactoryBean to set ProtocolVersion

clem0110:
https://jira.spring.io/browse/DATACASS-231

PR #49, opened 2016-04-10
Don't use deprecated methods of the Java driver.

stefanbirkner:


spring-cql/src/main/java/org/springframework/cassandra/core/converter/RowToListConverter.java
```diff
@@ -23,7 +24,8 @@
 
 		for (Definition def : cols.asList()) {
 			String name = def.getName();
-			list.add(row.isNull(name) ? null : def.getType().deserialize(row.getBytesUnsafe(name)));
+			list.add(row.isNull(name) ? null : def.getType().deserialize(
+					row.getBytesUnsafe(name), ProtocolVersion.NEWEST_SUPPORTED));
```
mp911de: We added `ProtocolVersion` with #41 so it would good to reuse it here.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/convert/RowReaderPropertyAccessor.java
```diff
@@ -51,7 +52,7 @@ public TypedValue read(EvaluationContext context, Object target, String name) {
 		}
 		DataType columnType = row.getColumnDefinitions().getType(name);
 		ByteBuffer bytes = row.getBytes(name);
-		Object object = columnType.deserialize(bytes);
+		Object object = columnType.deserialize(bytes, ProtocolVersion.NEWEST_SUPPORTED);
```
mp911de: We added `ProtocolVersion` with #41 so it would good to reuse it here.

mp911de:
Thanks @stefanbirkner for your PR. I added two comments to the PR. We're heading anyway towards the Cassandra 3 driver so we will have to inspect all method calls related to `ProtocolVersion`.

Can you please sign, if you haven't done already, the [CLA form](https://support.springsource.com/spring_committer_signup) for `Spring Data Access` with project lead `Oliver Gierke` and add your CLA Number as comment to this issue or [DATACASS-XXX](https://jira.spring.io/browse/DATACASS).

stefanbirkner:
CLA Number: 172420160413064110

spring-cql/src/main/java/org/springframework/cassandra/core/converter/RowToListConverter.java
stefanbirkner: Therefore this method has to know one of the objects `Cluster`, `Configuration`, `ProtocolOptions` or `ProtocolVersion` itself. This cannot be done without changing the method's signature or the class' constructor. In think this is the reason for using `ProtocolVersion.NEWEST_SUPPORTED` in commit e1a197170b94cf4df93d5b25c4db43a0c7f63282, too. I suggest to merge this commit like it is in order to avoid the exception. Replacing the `ProtocolVersion.NEWEST_SUPPORTED` should be a new issue that considers the other classes that use `ProtocolVersion`, too.

mp911de: Maybe that's the best option after all.

mp911de:
Thanks, that's merged with b3289af.

PR #55, opened 2016-05-11
DATACASS-219 - On startup CREATE TABLE from entities should only add 'if not exists'.

jxblum:
Fixes [DATACASS-219](https://jira.spring.io/browse/DATACASS-219).

Note, I also fixed the issue with `SchemaAction.RECREATE`.

### DIFF
--- a/spring-cql/src/test/java/org/springframework/cassandra/config/CassandraCqlSessionFactoryBeanUnitTests.java
+++ b/spring-cql/src/test/java/org/springframework/cassandra/config/CassandraCqlSessionFactoryBeanUnitTests.java
@@ -1,0 +1,349 @@
+/*
+ *  Copyright 2013-2016 the original author or authors
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cassandra.config;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.sameInstance;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.mockito.InOrder;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.springframework.cassandra.core.CqlOperations;
+
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.Session;
+
+/**
+ * The CassandraCqlSessionFactoryBeanUnitTests class is a test suite of test cases testing the contract
+ * and functionality of the {@link CassandraCqlSessionFactoryBean} class.
+ *
+ * @author John Blum
+ * @see org.junit.Rule
+ * @see org.junit.Test
+ * @see org.junit.rules.ExpectedException
+ * @see org.junit.runner.RunWith
+ * @see org.mockito.Mock
+ * @see org.mockito.Mockito
+ * @see org.mockito.runners.MockitoJUnitRunner
+ * @see org.springframework.cassandra.config.CassandraCqlSessionFactoryBean
+ * @since 1.5.0
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class CassandraCqlSessionFactoryBeanUnitTests {
+
+	@Rule
+	public ExpectedException exception = ExpectedException.none();
+
+	@Mock
+	private Cluster mockCluster;
+
+	@Mock
+	private Session mockSession;
+
+	private CassandraCqlSessionFactoryBean factoryBean;
+
+	protected <T> List<T> asList(T... array) {
+		return new ArrayList<T>(Arrays.asList(array));
+	}
+
+	protected void assertNonNullEmptyCollection(Collection<?> collection) {
+		assertThat(collection, is(notNullValue()));
+		assertThat(collection.isEmpty(), is(true));
+	}
+
+	@Before
+	public void setup() {
+		factoryBean = spy(new CassandraCqlSessionFactoryBean());
+	}
+
+	@Test
+	public void cassandraCqlSessionFactoryBeanIsSingleton() {
+		assertThat(factoryBean.isSingleton(), is(true));
+	}
+
+	@Test
+	public void objectTypeWhenSessionHasNotBeenInitializedIsSessionClass() {
+		assertThat(factoryBean.getObject(), is(nullValue()));
+		assertEquals(Session.class, factoryBean.<Session>getObjectType());
+	}
+
+	@Test
+	@SuppressWarnings("unchecked")
+	public void afterPropertiesSetInitializesSessionWithKeyspaceAndExecutesStartupScripts() throws Exception {
+		List<String> expectedStartupScripts = asList("/path/to/schema.cql", "/path/to/data.cql");
+
+		CqlOperations mockCqlOperations = mock(CqlOperations.class);
+
+		doReturn(mockSession).when(factoryBean).connect(eq("TestKeyspace"));
+		doReturn(mockCqlOperations).when(factoryBean).newCqlOperations(eq(mockSession));
+
+		factoryBean.setKeyspaceName("TestKeyspace");
+		factoryBean.setStartupScripts(expectedStartupScripts);
+
+		assertThat(factoryBean.getKeyspaceName(), is(equalTo("TestKeyspace")));
+		assertThat(factoryBean.getStartupScripts(), is(equalTo(expectedStartupScripts)));
+
+		factoryBean.afterPropertiesSet();
+
+		assertEquals(mockSession.getClass(), factoryBean.getObjectType());
+		assertThat(factoryBean.getObject(), is(equalTo(mockSession)));
+		assertThat(factoryBean.getSession(), is(equalTo(mockSession)));
+
+		InOrder inOrder = inOrder(factoryBean);
+
+		inOrder.verify(factoryBean, times(1)).connect(eq("TestKeyspace"));
+		inOrder.verify(factoryBean, times(1)).executeScripts(eq(expectedStartupScripts));
+		inOrder.verify(factoryBean, times(1)).newCqlOperations(eq(mockSession));
+		verify(mockCqlOperations, times(1)).execute(eq(expectedStartupScripts.get(0)));
+		verify(mockCqlOperations, times(1)).execute(eq(expectedStartupScripts.get(1)));
+	}
+
+	@Test
+	public void connectToSystemKeyspace() {
+		when(mockCluster.connect()).thenReturn(mockSession);
+
+		factoryBean.setCluster(mockCluster);
+
+		assertThat(factoryBean.connect(null), is(equalTo(mockSession)));
+
+		verify(mockCluster, times(1)).connect();
+		verify(mockCluster, never()).connect(anyString());
+	}
+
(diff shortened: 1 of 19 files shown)


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/mapping/BasicCassandraMappingContext.java
```diff
@@ -153,7 +153,7 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
 
 	@Override
 	public boolean usesTable(TableMetadata table) {
-		return entitySetsByTableName.containsKey(table.getName());
+		return entitySetsByTableName.containsKey(cqlId(table.getName()));
```
mp911de: 👍

spring-cql/src/test/java/org/springframework/cassandra/config/CassandraCqlSessionFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,349 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cassandra.config;
+
+import static org.hamcrest.Matchers.equalTo;
```
mp911de: Not a big deal, but let's stick to the [Spring Data Formatter settings](https://github.com/spring-projects/spring-data-cassandra/blob/master/etc/eclipse-formatting.xml). Static imports are imported with using wildcards then.

bildschirmfoto 2016-05-11 um 09 17 28

spring-cql/src/test/java/org/springframework/cassandra/config/CassandraCqlSessionFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,349 @@
...
+
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.Session;
+
+/**
+ * The CassandraCqlSessionFactoryBeanUnitTests class is a test suite of test cases testing the contract
+ * and functionality of the {@link CassandraCqlSessionFactoryBean} class.
+ *
+ * @author John Blum
+ * @see org.junit.Rule
```
mp911de: How about omitting `@see` tags in tests?

spring-cql/src/test/java/org/springframework/cassandra/config/CassandraCqlSessionFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,349 @@
...
+ * @see org.junit.Test
+ * @see org.junit.rules.ExpectedException
+ * @see org.junit.runner.RunWith
+ * @see org.mockito.Mock
+ * @see org.mockito.Mockito
+ * @see org.mockito.runners.MockitoJUnitRunner
+ * @see org.springframework.cassandra.config.CassandraCqlSessionFactoryBean
+ * @since 1.5.0
```
mp911de: Please add `@see DATACASS-xxx` ticket references to the tests

spring-cql/src/test/java/org/springframework/cassandra/core/util/CollectionUtilsUnitTests.java
```diff
@@ -0,0 +1,143 @@
...
+		assertThat(array, is(notNullValue()));
+		assertThat(array.length, is(equalTo(0)));
+	}
+
+	void assertNonNullEmptyCollection(Collection<?> collection) {
+		assertThat(collection, is(notNullValue()));
+		assertThat(collection.isEmpty(), is(true));
+	}
+
+	<T> Iterable<T> newIterable(final T... elements) {
```
mp911de: Please help me understand why `newIterable` is needed. Assumed `Arrays.asList(…)` or `new ArrayList<>(…)` was sufficient for the performed tests.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CassandraSessionFactoryBean.java
```diff
@@ -13,109 +13,153 @@
...
+
+	private SchemaAction schemaAction = SchemaAction.NONE;
 
 	@Override
 	public void afterPropertiesSet() throws Exception {
-
 		super.afterPropertiesSet();
 
-		Assert.notNull(converter);
+		Assert.state(converter != null, "Converter must not be null");
```
mp911de: I guess there's no consistent rule for exceptions in `afterPropertiesSet` - even the core framework throws `IllegalArgument`/`NullPointer`/other exceptions. Was just wondering, all good here.

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/config/CassandraSessionFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,342 @@
...
+import com.datastax.driver.core.Metadata;
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.TableMetadata;
+
+/**
+ * The CassandraSessionFactoryBeanUnitTests class is a test suite of test cases testing the contract and functionality
+ * of the {@link CassandraSessionFactoryBean} class.
+ *
+ * @author John Blum
+ * @see org.junit.Rule
```
mp911de: How about omitting `@see (classname)` tags in tests?

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/config/CassandraSessionFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,342 @@
...
+ * @see org.junit.Test
+ * @see org.junit.rules.ExpectedException
+ * @see org.junit.runner.RunWith
+ * @see org.mockito.Mock
+ * @see org.mockito.Mockito
+ * @see org.mockito.runners.MockitoJUnitRunner
+ * @see org.springframework.data.cassandra.config.CassandraSessionFactoryBean
+ * @since 1.5.0
```
mp911de: Please add `@see DATACASS-xxx` ticket references to the tests

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/integration/config/SchemaActionIntegrationTests.java
```diff
@@ -0,0 +1,286 @@
...
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.TableMetadata;
+import com.datastax.driver.core.exceptions.AlreadyExistsException;
+
+/**
+ * The SchemaActionIntegrationTests class is a test suite of test cases testing the contract and behavior
+ * of various {@link SchemaAction}s on startup of a Spring configured, Cassandra application client.
+ *
+ * @author John Blum
+ * @see org.junit.Test
```
mp911de: How about omitting `@see (classname)` tags in tests?

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/integration/config/SchemaActionIntegrationTests.java
```diff
@@ -0,0 +1,286 @@
...
+ * @see org.junit.Rule
+ * @see org.junit.rules.ExpectedException
+ * @see org.springframework.cassandra.test.integration.AbstractEmbeddedCassandraIntegrationTest
+ * @see org.springframework.cassandra.test.integration.KeyspaceRule
+ * @see org.springframework.data.cassandra.config.CassandraSessionFactoryBean
+ * @see org.springframework.data.cassandra.config.SchemaAction
+ * @see org.springframework.data.cassandra.config.java.AbstractCassandraConfiguration
+ * @since 1.0.0
```
mp911de: `1.5.0`

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/integration/config/SchemaActionIntegrationTests.java
```diff
@@ -0,0 +1,286 @@
...
+ */
+public class SchemaActionIntegrationTests extends AbstractEmbeddedCassandraIntegrationTest {
```
mp911de: Please add `@see DATACASS-xxx` ticket references to the tests

mp911de:
Like the direction where Spring Data Cassandra is heading to. Please apply [Spring Data Formatting](https://github.com/spring-projects/spring-data-cassandra/blob/master/etc/eclipse-formatting.xml) to the code. Can support you on setting it up with IntelliJ. Added some minor comments.

spring-cql/src/test/java/org/springframework/cassandra/config/CassandraCqlSessionFactoryBeanUnitTests.java
jxblum: Will do.

Personally, I think it was a mistake for Java to ever allow wildcard imports.  It is always better to be explicit about what exactly is being imported (a dependency, used by a class), static or not.  It is a simple matter to include (auto) import classes from within an IDE so using wildcards has less value today then it did in the past.  Using wildcards also makes it more difficult, especially outside an IDE to determine where a conflict originates.  No where is this more apparent then with Mockito and Hamcrest Matchers.

spring-cql/src/test/java/org/springframework/cassandra/config/CassandraCqlSessionFactoryBeanUnitTests.java
jxblum: Right; I suppose these are not all that useful in tests eh, ;-)

spring-cql/src/test/java/org/springframework/cassandra/config/CassandraCqlSessionFactoryBeanUnitTests.java
jxblum: Will do.  Though, I do think we should be using the proper syntax for the `@see` tag when referencing external references.

spring-cql/src/test/java/org/springframework/cassandra/core/util/CollectionUtilsUnitTests.java
jxblum: Yeah not really needed.  Just wanted a pure `Iterable` object that was not based on some `Collection` (e.g. `List`) for true unit testing.  But it is of little consequence.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CassandraSessionFactoryBean.java
jxblum: I think the Exception is important in this case (well, in most cases) as it communicates either the state of the object/system (`IllegalStateException`), the validity of the values passed in (`IllegalArgumentException`) or whether the programmer (caller) made a mistake using the API (`NullPointerException`, a programmatical `Exception`).

In this, case the `IllegalStateException` represents that the object is not in a proper state to finish post initialization activities, therefore, the `IllegalStateException` is the appropriate `Exception`.

However, what is probably not appropriate is the message.  The `Exception` message should rather read...

> Converter was not properly initialized

Will change.

Side note...

I do think there is an excessive number of inappropriate `Assert.notNull(object)`, `Assert.isTrue(condition)`, etc in the Spring Data codebase that should be including a **message**.

One of the reasons for this is, the Spring `Assert` utility class uses a generic message when one is not given, for example... `[Assert.notNull(object)](https://github.com/spring-projects/spring-framework/blob/v4.2.6.RELEASE/spring-core/src/main/java/org/springframework/util/Assert.java#L126)`.

The problem is, if a constructor or method with multiple arguments (for [example](https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/mapping/model/AbstractPersistentProperty.java#L61-L62)) asserts not null on those arguments without a message, such as...

```
public Object someMethod(Object argument1, Object argument2, ...) {
    Assert.notNull(argument1);
    Assert.notNull(argument2);

    ...
}
```

Then, the caller has no idea which argument was null, given the user no context and helpful debug information.

Food for thought.

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/config/CassandraSessionFactoryBeanUnitTests.java
jxblum: Will do, thx.

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/config/CassandraSessionFactoryBeanUnitTests.java
jxblum: Will do.

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/integration/config/SchemaActionIntegrationTests.java
jxblum: Will do.

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/integration/config/SchemaActionIntegrationTests.java
jxblum: Good catch, thank you.

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/integration/config/SchemaActionIntegrationTests.java
jxblum: Will do.

jxblum:
Added the [Spring Data Formatting](https://github.com/spring-projects/spring-data-cassandra/blob/master/etc/eclipse-formatting.xml) to my IDE for the Spring Data projects.  Thanks for the tip.

jxblum:
Thank you @mp911de for the code review.  Good tips.  Cheers.

PR #57, opened 2016-05-20
fix whitespace of pom files

atoulme:
This is a simple set of whitespace changes to format all pom.xml files to use 2 spaces for indentation.


gregturn:
What is the basis for using  2-spaced system? None of the other Spring projects do that.

atoulme:
You're right @gregturn. I am changing to tabs to match spring-data-jpa.

mp911de:
Closing this PR as we don't do PR's for [formatting-related changes](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc#quickstart).

PR #59, opened 2016-05-21
DATACASS-250: Support lightweight transactions (if [not] exists)

atoulme:
Add updateIfExists and insertIfNotExists to CassandraOperations.
Implemented the methods in CassandraTemplate.
Added some tests to existing integration tests.


mp911de:
This PR introduces a lot of methods for just one flag for `INSERT`/`UPDATE`. It requires a bit more thought as the current approach does not scale well with extending the API. I'm not disregarding your work. Adding several methods for query variations pollutes the API and is not a clean approach.

PR #60, opened 2016-05-24
DATACASS-156 - CqlOperations' ingest method signatures taking WriteOp…

mbazos:
Submitting a PR as this caused some confusion on our team, we were using ingest setting the TTL on the WriteOptions but wasn't actually doing anything.

As an alternative when using ingest the cql passed in has the TTL in the query.


jxblum:
Will look at this when #61 has been reviewed/merged.

mp911de:
@mbazos thanks for your pull request and sorry for the long radio silence.

We decided to remove `ingest` in future versions of Spring Data for Apache Cassandra (for 2.0.x) because they create more harm than value. Therefore, this PR is no longer applicable.

PR #61, opened 2016-05-24
DATACASS-286 - Log all CQL queries executed with CqlTemplate.

jxblum:
In addition to adding logging for all CQL statements in the template, I also spent additional effort/time to refactor this class (and interface) to simplify the logic and make the (similar) methods more consistent in behavior.

While much has been changed, even more refactoring is required to bring this class up to standard.  I started a minimal set of unit tests mainly covering the areas that significantly changed (i.e. the method contract is now (somewhat) different, e.g. throwing a different Exception).

Some of the change may seem incorrect/risky, such as [removing the null check](https://github.com/spring-projects/spring-data-cassandra/blob/1.4.1.RELEASE/spring-cql/src/main/java/org/springframework/cassandra/core/CqlTemplate.java#L545) around the listener in the `doExecuteAsync(..)` method.  However, even null checking was [not done consistently](https://github.com/spring-projects/spring-data-cassandra/blob/1.4.1.RELEASE/spring-cql/src/main/java/org/springframework/cassandra/core/CqlTemplate.java#L#L357-L362) in the template logic.

My goals were to simplify, enforce consistency and remove redundancy, in places.  Still more work needs to be done, but this is a start.


mp911de:
That's merged with 1d1ba3d.

PR #65, opened 2016-06-14
DATACASS-297 - Add support for streaming queries to select entities.

mp911de:
We now support java.util.stream.Stream as return type on query methods. Streaming queries wrap the iterator returned by Cassandra's ResultSet. Elements are fetched in batches (see QueryOptions that can be set in CassandraCqlClusterFactoryBean) and processed element by element instead of reading all results. Streaming queries are supported only for select queries.

----

Related ticket: [DATACASS-297](https://jira.spring.io/browse/DATACASS-297)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-cassandra-parent</artifactId>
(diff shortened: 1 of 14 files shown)


jxblum:
PR #65 [merged](https://github.com/spring-projects/spring-data-cassandra/commit/1b4eec9dd40141cd6359f6d6e8c94b93691eee3c).

PR #66, opened 2016-06-17
DATACASS-298 - Add missing PoolingOptions to the XML namespace as well as the PoolingOptionsFactoryBean.

jxblum:


spring-cql/src/main/java/org/springframework/cassandra/config/PoolingOptionsFactoryBean.java
```diff
@@ -15,226 +15,455 @@
...
-		if (remoteMaxConnections != null) {
-			poolingOptions.setMaxConnectionsPerHost(HostDistance.REMOTE, remoteMaxConnections);
-		}
+		HostDistancePoolingOptions local = LocalHostDistancePoolingOptions.create(localCoreConnections,
+			localMaxConnections, localMaxSimultaneousRequests, localMinSimultaneousRequests);
 
-		if (remoteCoreConnections != null) {
-			poolingOptions.setCoreConnectionsPerHost(HostDistance.REMOTE, remoteCoreConnections);
-		}
+		local.setMaxConnectionsPerHost(poolingOptions);
```
mp911de: Maybe add another method to extract the same calls for `HostDistancePoolingOptions`.

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,229 @@
...
+		poolingOptionsFactoryBean.setPoolTimeoutMilliseconds(180);
+
+		assertThat(poolingOptionsFactoryBean.getObject(), is(nullValue(PoolingOptions.class)));
+
+		poolingOptionsFactoryBean.afterPropertiesSet();
+
+		assertThat(poolingOptionsFactoryBean.getObject(), is(sameInstance(mockPoolingOptions)));
+		assertThat(poolingOptionsFactoryBean.getObjectType(), is(equalTo((Class) mockPoolingOptions.getClass())));
+
+		InOrder inOrder = inOrder(mockPoolingOptions);
```
mp911de: `InOrder` tends to cause a lot of maintenance once tests are in place. We should have some tests which take real XML config and create the `PoolingOptions`. Who knows what happens inside of `PoolingOptions` in the future.

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,229 @@
...
+		poolingOptionsFactoryBean.setPoolTimeoutMilliseconds(120);
+
+		assertThat(poolingOptionsFactoryBean.getObject(), is(nullValue(PoolingOptions.class)));
+
+		poolingOptionsFactoryBean.afterPropertiesSet();
+
+		assertThat(poolingOptionsFactoryBean.getObject(), is(sameInstance(mockPoolingOptions)));
+		assertThat(poolingOptionsFactoryBean.getObjectType(), is(equalTo((Class) mockPoolingOptions.getClass())));
+
+		InOrder inOrder = inOrder(mockPoolingOptions);
```
mp911de: see above w.r.t `InOrder`.

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,229 @@
...
+		inOrder.verify(mockPoolingOptions, times(1)).setMaxRequestsPerConnection(eq(HostDistance.LOCAL), eq(50));
+		inOrder.verify(mockPoolingOptions, times(1)).setNewConnectionThreshold(eq(HostDistance.LOCAL), eq(5));
+		verify(mockPoolingOptions, never()).setInitializationExecutor(any(Executor.class));
+		verify(mockPoolingOptions, never()).setCoreConnectionsPerHost(eq(HostDistance.REMOTE), anyInt());
+		verify(mockPoolingOptions, never()).setMaxConnectionsPerHost(eq(HostDistance.REMOTE), anyInt());
+		verify(mockPoolingOptions, never()).setMaxRequestsPerConnection(eq(HostDistance.REMOTE), anyInt());
+		verify(mockPoolingOptions, never()).setNewConnectionThreshold(eq(HostDistance.REMOTE), anyInt());
+	}
+
+	@Test
```
mp911de: Add ticket reference

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,229 @@
...
+		assertThat(poolingOptionsFactoryBean.getLocalMaxSimultaneousRequests(), is(equalTo(200)));
+		assertThat(poolingOptionsFactoryBean.getLocalMinSimultaneousRequests(), is(equalTo(100)));
+		assertThat(poolingOptionsFactoryBean.getPoolTimeoutMilliseconds(), is(equalTo(300)));
+		assertThat(poolingOptionsFactoryBean.getRemoteCoreConnections(), is(equalTo(25)));
+		assertThat(poolingOptionsFactoryBean.getRemoteMaxConnections(), is(equalTo(250)));
+		assertThat(poolingOptionsFactoryBean.getRemoteMaxSimultaneousRequests(), is(equalTo(100)));
+		assertThat(poolingOptionsFactoryBean.getRemoteMinSimultaneousRequests(), is(equalTo(50)));
+	}
+
+	@Test
```
mp911de: Add ticket reference

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,229 @@
...
+	private PoolingOptions mockPoolingOptions;
+
+	private PoolingOptionsFactoryBean poolingOptionsFactoryBean;
+
+	@Before
+	public void setup() {
+		poolingOptionsFactoryBean = new PoolingOptionsFactoryBean();
+	}
+
+	@Test
```
mp911de: Add ticket reference

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,229 @@
...
+	public void isSingletonIsTrue() {
+		assertThat(poolingOptionsFactoryBean.isSingleton(), is(true));
+	}
+
+	@Test
```
mp911de: Add ticket reference

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,229 @@
...
+	public void getObjectReturnsNullWhenNotInitialized() throws Exception {
+		assertThat(poolingOptionsFactoryBean.getObject(), is(nullValue(PoolingOptions.class)));
+	}
+
+	@Test
```
mp911de: Add ticket reference

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
```diff
@@ -0,0 +1,229 @@
...
+
+	/**
+	 * This particular test case is technically an integration test since it uses an actual instance of
+	 * a DataStax Java driver class type... {@link PoolingOptions}!
+	 *
+	 * The max values should be set before setting core values. Otherwise the core values will be compared with the
+	 * default max values which is 8. Same for other min-max properties pairs. This test checks the same.
+	 *
+	 * @throws Exception Any unhandled scenarios will result in a test failure.
+	 * @see <a href="https://jira.spring.io/browse/DATACASS-176">DATACASS-176</a>
```
mp911de: We should omit the host/path part as URL's tend to change

spring-cql/src/test/java/org/springframework/cassandra/config/xml/CassandraCqlClusterParserUnitTests.java
```diff
@@ -0,0 +1,403 @@
...
+
+/**
+ * Test suite of Unit tests testing the contract and functionality of the {@link CassandraCqlClusterParser}.
+ *
+ * @author John Blum
+ * @see org.springframework.cassandra.config.xml.CassandraCqlClusterParser
+ * @since 1.5.0
+ */
+// TODO add more tests!
+@RunWith(MockitoJUnitRunner.class)
```
mp911de: A set of tests using real XML config seems more appealing to me. Less stubbing/setup and so tests stay more maintainable.

spring-data-cassandra/src/main/resources/org/springframework/data/cassandra/config/spring-cassandra-1.0.xsd
```diff
@@ -1,12 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
mp911de: There's also `spring-cql-1.0.xsd` in `spring-cql` which declares nearly the same elements. We should change the other one as well until we get to the point to keep only one of the two.

mp911de:
Did a first pass. Impressive changes. Added some comments to the PR.

We should adopt `spring-cql-1.0.xsd` to keep it in sync with `spring-cassandra-1.0.xsd`. 

We should also have some tests that perform an integration test for `PoolingOptions` for both namespaces `http://www.springframework.org/schema/cql` and `http://www.springframework.org/schema/data/cassandra`. I see the advantage in less stubbing (less maintenance in tests), and we test against a real `PoolingOptions` instance. Who knows what happens in future versions to `PoolingOptions`. This way we have a direct indicator if something breaks.

spring-cql/src/main/java/org/springframework/cassandra/config/PoolingOptionsFactoryBean.java
jxblum: Done

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
jxblum: Yeah, I contemplated this a lot.

Personally, I feel it is overkill to express this in XML and actually start an actual Cassandra instance (embedded or not), just to verify the DataStax `PoolingOptions` instance.  Using actual XML is good for actual test case(s) involving and testing the parsers, but even then I question whether an actual instance of (embedded?) Cassandra is needed for parser tests either.

SDG makes use of mocking, even at the integration test level, with the use of a _Spring_ [ApplicationContextInitializer](https://github.com/jxblum/spring-data-gemfire/blob/master/src/test/java/org/springframework/data/gemfire/test/GemfireTestApplicationContextInitializer.java) and the help of the [GemfireTestBeanPostProcessor](https://github.com/jxblum/spring-data-gemfire/blob/master/src/test/java/org/springframework/data/gemfire/test/GemfireTestBeanPostProcessor.java) which creates mock versions of the SDG `FactoryBeans` to create mock GemFire components (e.g. Cache) instead of actual GemFire components (i.e. `GemFireCacheImpl`).

This is especially nice for testing the XML namespace elements where it is only important that the configuration meta-data used to configure and initialize the underlying data store components actually occurred and was proper (according to the API), not whether the actual underlying data store components behave correctly according to the configuration.

All of this is to say that I would eventually like to employ similar testing patterns and techniques to SD Cassandra to keep the test suite as light and fast as possible, thus relying more on "unit" tests over "integration" tests when the functionality/behavior can adequately be covered in unit tests where possible.

In this particular case, that is exactly what the [existing test case method](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTest.java#L45-L68) was testing, an actual instance of the `PoolingOptions` class in order to address bug [DATACASS-176](https://jira.spring.io/browse/DATACASS-176) (see also [here](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTest.java#L43) along with the [comment](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTest.java#L39-L40)).

I had half a thought of rewriting this test case method since...

1. What is [this](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTest.java#L59-L66) really testing? (argh)

2. I mean, if you are going to call [afterPropertiesSet()](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTest.java#L57) without mocks, you might as well test the actual `PoolingOptions` instance created, which is exactly how I modified this test.

I thought about adding order verification here, but I decided not.

In addition, I have cleaned up all instances of my order verification use in my additional test cases in the updated PR.

Let me know if you have any additional thoughts on this.  Thx.

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
jxblum: Cleaned up; thx.

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
jxblum: I switched to using a `Spy` as well.

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
jxblum: Done

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
jxblum: Done

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
jxblum: Done (and done, for DATACASS-176 as well).

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
jxblum: The other test case methods: `getObjectReturnsNullWhenNotInitialized`, `getObjectTypeReturnsPoolingOptionsClassWhenNotInitialized` and `isSingletonIsTrue` have nothing to do with DATACASS-298.  They were added because the overall `PoolingOptionsFactoryBean` class was missing (and still is to a certain extent) adequate unit test coverage.

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
jxblum: The other test case methods: `getObjectReturnsNullWhenNotInitialized`, `getObjectTypeReturnsPoolingOptionsClassWhenNotInitialized` and `isSingletonIsTrue` have nothing to do with [DATACASS-298](https://jira.spring.io/browse/DATACASS-298).  They were added because the overall `PoolingOptionsFactoryBean` class was missing (and still is to a certain extent) adequate unit test coverage.

spring-cql/src/test/java/org/springframework/cassandra/config/xml/CassandraCqlClusterParserUnitTests.java
jxblum: See my previous comments above.  Agreed except for the part of actually starting an actual (embedded) Cassandra instance, which is quite overkill, and currently, SD Cassandra does not have good mocking support, such as say SDG's test suite.

spring-data-cassandra/src/main/resources/org/springframework/data/cassandra/config/spring-cassandra-1.0.xsd
jxblum: Shoot, I did not realize that; darn it.  This is both unfortunate and redundant, and `spring-cql-1.0.xsd` should probably be **deprecated** since it is a subset of the `spring-cassandra-1.0` XSD.

Updated.

spring-cql/src/test/java/org/springframework/cassandra/config/PoolingOptionsFactoryBeanUnitTests.java
jxblum: Are there options in the `Javadoc` Maven plugin to, perhaps, set `baseUrl` and/or `path` that could serve a placeholder for properties used in source?  (yuck) Though, saying that seems equally flawed since branch source will potentially become out-of-date if the URL ever does change.

Personally, I am not a big fan of improper syntax in `Javadoc`, so SD's use of the `@see` tags (i.e. @see DATACASS-xyz`) is rather broken.  I'd rather see no tag then an improper one.

Perhaps we can refer to JIRA tickets in another way, such as...

```
/**
 * Refer to DATACASS-xyz.
 */
```

jxblum:
Regarding your "summary" comment.  Thank you 👍, and thank you for reviewing.

I have updated existing test classes: `XmlConfgIntegrationTests` and `PropertyPlaceholderNamespaceCreatingXmlConfigIntegrationTests`.

In so doing, I found (yet) another bug.  Since both the [XmlConfigIntegrationTests](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-cql/src/test/java/org/springframework/cassandra/test/integration/config/xml/XmlConfigIntegrationTests.java#L56-L59) and [PropertyPlaceholderNamespaceCreatingXmlConfigIntegrationTests](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-cql/src/test/java/org/springframework/cassandra/test/integration/config/xml/PropertyPlaceholderNamespaceCreatingXmlConfigIntegrationTests.java#L46-L53) did not really very anything (!), really, I amped up the test coverage and found [this](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-cql/src/test/resources/org/springframework/cassandra/test/integration/config/xml/XmlConfigIntegrationTests-context.xml#L21) corresponding to [this](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-cql/src/main/java/org/springframework/cassandra/config/xml/CassandraCqlClusterParser.java#L279).  Needless to say it did not get the proper setting for 'read-timeout'.

Anyway, I have fixed the issue and will be including that in the changes for [DATACASS-298](https://jira.spring.io/browse/DATACASS-298) (rather than open another bug).

jxblum:
Aiy... and a similar problem with `connection-time-millis` defined in the `socketOptionsType` XSD complexType element, yet the `CassandraCqlClusterParser`

mp911de:
Thanks a lot. That's merged with 3d9835d and polished with 56c46f9.

PR #72, opened 2016-06-28
DATACASS-182 - Add ability to update/store null into fields.

mp911de:
We now support the removal of columns for properties with `null` values by allowing null values in `Insert` and `Update` statements written by `MappingCassandraConverter`. Objects are stored with all persistent properties. Insert and update perform no longer conditional inserting of non-null values but take all values into account.

----

Related ticket: [DATACASS-182](https://jira.spring.io/browse/DATACASS-182)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-cassandra-parent</artifactId>
-	<version>1.5.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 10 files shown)


jxblum:
PR #72 merged to `master` in commit [e10c717](https://github.com/jxblum/spring-data-cassandra/commit/e10c71722dec1f1caea5b71e2133ec23e79762fb) for the _Spring Data Cassandra_ **1.5 M1 (Ingalls)** release.

PR #73, opened 2016-07-01
DATACASS-206, DATACASS-308 Fix delete using custom column names and NullPointerException related properties using the MapId mechanism

mp911de:
We now use the specified column name from the entity annotation to construct the where clause in a delete statement.

We also now reject unknown/misspelled property names when using `MapId`. The Id conversion and extraction now takes place in the Cassandra Converter. The Id handling code in `CassandraTemplate` is deprecated.

Also, some polishing to in that context to reorder fields in MappingCassandraConverter according to their visibility. Remove code duplicates. Improve JavaDoc in CassandraOperations. Align entity class parameter names in CassandraTemplate. Enhance JavaDoc for built statement factory methods. Add guards to method arguments.

----

Related tickets: [DATACASS-206](https://jira.spring.io/browse/DATACASS-206) and [DATACASS-308](https://jira.spring.io/browse/DATACASS-308).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-cassandra-parent</artifactId>
-	<version>1.5.0.BUILD-SNAPSHOT</version>
+	<version>1.5.0.DATACASS-206-SNAPSHOT</version>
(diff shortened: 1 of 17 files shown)


mp911de:
Closing as it was merged.

PR #81, opened 2016-07-28
DATACASS-328 - Revise QueryOptions and WriteOptions.

mp911de:
Deprecate our `org.springframework.cassandra.core.ConsistencyLevel` enum. Having an own consistency level type leads to confusion and it's always behind the driver. We don't want to maintain that type, so we decided to deprecate the own type and use the driver consistency levels.

We allow now the use of the driver retry policies aside of our consistency level enumeration. For most cases, our enumeration is the simpler approach. Some retry policies (`IdempotenceAwareRetryPolicy`, `LoggingRetryPolicy`) require further configuration and cannot be applied with just using a static enum value. We now support ReadTimeout, FetchSize, and Tracing via `QueryOptions` and `WriteOptions` and provide builders for `QueryOptions` and `WriteOptions`.

----

Related ticket: [DATACASS-328](https://jira.spring.io/browse/DATACASS-328)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-cassandra-parent</artifactId>
-	<version>1.5.0.BUILD-SNAPSHOT</version>
+	<version>1.5.0.DATACASS-328-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data Cassandra</name>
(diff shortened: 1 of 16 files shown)


jxblum:
PR #81 [merged](https://github.com/spring-projects/spring-data-cassandra/commit/fd8e4305923c46286bd8b618ead8edb76716b9cd) to SDC `master` for the **1.5 RC1** release.

PR #86, opened 2016-10-21
DATACASS-335 - Add support for reactive data access.

mp911de:
We now support reactive data access with Spring Data Cassandra by adopting Datastax' asynchronous driver.

`ReactiveCqlTemplate` and `ReactiveCassandraTemplate` use [Project Reactor](http://projectreactor.io/) wrapper types `Mono` and `Flux` to implement Template API and repository support. Reactive template supports common operations such as:
- Query/Execution methods for static CQL and prepared statements
- Insert/Save/Update/Delete methods
- Exists and Count projections
- Reactive Callback methods

``` java
Person person = new Person("Dave", 25);

template.insert(person) //
    .flatMap(p -> template.update(new Person("Sven", 25))) //
    .flatMap(p -> template.selectOneById(person.getId(), Person.class)) //
    .subscribeWith(TestSubscriber.create()) //
    .await() //
    .assertValuesWith(result -> {
        assertThat(result.getFirstName(), is(equalTo("Sven")));
    });
```

Reactive Repository support is built on top of ReactiveCassandraTemplate using ReactiveCassandraRepository as the store-specific base repository. Reactive repositories are enabled by using`@EnableReactiveCassandraRepositories` on a `@Configuration` class to opt-in for reactive support. Reactive repositories can be composed of a reactive base interface such as
- `ReactiveCrudRepository`
- `ReactiveSortingRepository`
- `RxJava1CrudRepository`
- `RxJava1SortingRepository`

and are identified as reactive repository if one method uses a reactive wrapper type (such as `Flux` or `Observable`). If a reactive repository is discovered, it's not implemented by the blocking repository support but with the reactive repository factory. Blocking methods are not (yet) synchronized when using a reactive repository so each repository method must use a reactive wrapper result type. Reactive repository support with Spring Data allows using RxJava1 and Project Reactor types to declare repository methods. Reactive wrapper types are internally converted so the composition library choice on repository level is left up to the user.

There's feature parity between Reactive Cassandra repository support and blocking repository support.

Feature overview:
- Query Methods using String queries and Query Derivation
- Projections

``` java
@Configuration
@EnableReactiveCassandraRepositories
class ApplicationConfig extends AbstractReactiveCassandraConfiguration {

  @Override
  protected String getKeyspaceName() {
    return "mykeyspace";
  }

  @Override
  protected String getEntityBasePackages() {
    return new String[] {"com.springdata.cassandra"};
  }
}

public interface PersonRepository extends ReactiveSortingRepository<Person, String> {

[...]

PR #87, opened 2016-10-25
DATACASS-292 - Provide revised synchronous and asynchronous CQL and Cassandra templates.

mp911de:
We now provide revised CQL and Cassandra templates as central classes to interact with CQL and Cassandra with object mapping. Previously, synchronous and asynchronous methods were exposed inside the same interfaces that made it hard to chose the right method.

The new Template API consists of:
- `CqlTemplate`
- `AsyncCqlTemplate`
- `CassandraTemplate`
- `AsyncCassandraTemplate`

`CassandraTemplate` and `AsyncCassandraTemplate` reuse `CqlTemplate` and `AsyncCqlTemplate` instead of extending from these. This is, to not mix methods using conversion/object mapping with lower level CQL execution methods.

`AsyncCqlTemplate` and `AsyncCassandraTemplate` are all new and benefit from `ListenableFuture` as synchronization aid. They no longer rely on various callback-interfaces. 

`CassandraTemplate` and `AsyncCassandraTemplate` no longer provide insert/update/delete methods accepting a collection of items. Use `CassandraBatchOperations` for atomic batches to group operations.

---

Related ticket: [DATACASS-292](https://jira.spring.io/browse/DATACASS-292)
Depends on: #86


jxblum:
Hi @mp911de -

I think we agreed that this should be part of `2.0.x` and **not** the current 1.x line given this PR introduces interface breaking changes.  Could you please rebase this PR based on the latest `2.0.x` branch before I review.

Thanks.

jxblum:
Reviewed and merged to `2.0.x` in revisions [8af2388](https://github.com/spring-projects/spring-data-cassandra/commit/8af23880e97fbf17d68a605771d7d5cae9bd82e9) and [542ce24](https://github.com/spring-projects/spring-data- cassandra/commit/542ce241736b735d4e4553b75470b0ab5fd28259) for the _Spring Data Cassandra_ **2.0.0.M1 (Kay)** release.

PR #93, opened 2017-01-20
DATACASS-330 - Add SessionFactory to CqlTemplate and AsyncCqlTemplate with routing capabilities.

mp911de:
We now support SessionFactory to obtain Cassandra `Session`s on a per-request basis. CassandraAccessor is configured primarily with a SessionFactory now, the existing initialization configures a DefaultSessionFactory that returns the initially given Session instance.

Sessions should not be acquired directly by `getSession` but inside a callback-block so it's guaranteed to operate on the same session within a particular operation. That's especially relevant when preparing and executing prepared statements.

`PreparedStatementCallback` was changed in a breaking way as it now accepts additionally a `Session` to retain the session context.

Move also reactive session classes into the `o.s.c.c.session` package.

---

Related ticket: [DATACASS-330](https://jira.spring.io/browse/DATACASS-330)


mgeiss:
Hey @mp911de,

this looks pretty good. It would be nice if it would be possible to add new sessions dynamically during run time, maybe be providing a protected method addTargetSession because we are adding Sessions only when needed. We will create new tenants dynamically without redeploying the component, so we figure out the details using an internal meta structure, and create the Session dynamically.

BTW, we do the same with the JPA module, simply using our own maps, overriding 'determineTargetDataSource' too. But it would be more cool to keep the logic for retrieving data source internal, and allowing to add data sources on the fly.

mp911de:
@mgeiss Thanks for your feedback. I added `SessionFactoryLookup` to our API and a test showing how to add new `SessionFactory` instances on the fly.

baasiq:
@mp911de thanks. looks great ; o) and I will start using it right away ... guess it'll be part of the M2 release for 2.0.0. ; o)

PR #100, opened 2017-03-01
DATACASS406 Support nested @UserDefinedType for schema creation

schauder:


mp911de:
That's merged and polished.

PR #103, opened 2017-05-22
JRA-167 #time 2h

eighty4:
Supprting @Embedded properties in entities to serialize/deserialize as json.

I took a stab at 167 and wanted to get feedback on the approach and additional test cases that should be done. I am directly mapping to JSON with an ObjectMapper in MappingCassandraConverter.

Would this better be done by populating ConversionService with Converters to handle this? Should the ObjectMapper be autowired instead and have a default Impl provided via a @Configuration if there isn't one in the application already or best to use a separate instance configured/constructed just for this? Is there another approach I overlooked?

I expected some changes to come from the PR, so not considering this ready to merge as is.

- [ ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [ ] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATACASS).
- [ ] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [ ] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/spring-data-cassandra/pom.xml
+++ b/spring-data-cassandra/pom.xml
@@ -127,22 +127,20 @@
 		</dependency>
 
 		<dependency>
-			<groupId>org.xerial.snappy</groupId>
-			<artifactId>snappy-java</artifactId>
-			<scope>test</scope>
-		</dependency>
-
-		<dependency>
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-core</artifactId>
 			<version>${jackson}</version>
-			<scope>test</scope>
 		</dependency>
 
 		<dependency>
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
 			<version>${jackson}</version>
+		</dependency>
+
+		<dependency>
+			<groupId>org.xerial.snappy</groupId>
+			<artifactId>snappy-java</artifactId>
 			<scope>test</scope>
 		</dependency>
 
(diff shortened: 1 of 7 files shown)


pivotal-issuemaster:
@eighty4 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-cassandra&pullRequestId=103)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-cassandra&pullRequestId=103) to manually synchronize the status of this Pull Request.

mp911de:
Thanks @eighty4 for submitting a pull request. 

You're right on track with `ConversionService` (see [custom converters](http://docs.spring.io/spring-data/cassandra/docs/current/reference/html/#cassandra.custom-converters)) which is the indended place for object to JSON or object to custom `UDTValue`/`TupleValue` serialization.

JSON mapping is always subject to some sort of configuration, especially with Jackson its broad support for extensions. Going out of the users' way and providing an extension point with custom conversion is the best option for us right now. Users should have all they need to easily implement Object to JSON serialization. 

We support right now two flavors of embeddable types: Compound primary keys and mapped UDTs. [DATACASS-167](https://jira.spring.io/browse/DATACASS-167) is a rather old ticket and it reads as support for the custom Cassandra data type.

We were considering to add embeddable types that map directly to columns, similar to what JPA provides with `@Embedded`/`@Embeddable`. Users could map a type to a subset of the table columns where the mapped type is embedded within the root table declaration. Right now, we're heading towards our GA release. Embeddable types have not the highest priority, that's why we don't actively work on the ticket.

mp911de:
Closing this PR because we want to solve embeddable types differently.

PR #111, opened 2017-07-19
DATACASS-213, DATACASS-306 - Create secondary indexes for annotated entities

mp911de:
We now create secondary indexes for annotated entities via `@Indexed` and `@SASI`. Index creation is part of schema creation that is executed after Session initialization and table creation.  
We support plain secondary indexes and key/value/entry indexes for map columns. Index creation is useful for rapid development but should not be used in large setups or at least with care to not impact performance in a negative way.  

```java
@Table
public class Person {
    
    @Id
    private String key;
    
    @Indexed("name_index")
    private String name;
    
    private Map<@Indexed String, String> keys;

    @SASI 
    private String names;

    @SASI 
    @StandardAnalyzed(value = "de",
                enableStemming = true, normalization = Normalization.UPPERCASE,
                skipStopWords = true) 
    private String profession;

    @SASI 
    @NonTokenizingAnalyzed
    private String country;
}
```
---

Related tickets:
* [DATACASS-213](https://jira.spring.io/browse/DATACASS-213)
* [DATACASS-306](https://jira.spring.io/browse/DATACASS-306)


jxblum:
Reviewed, polished and merged to `master` for the _Spring Data Cassandra_ **Kay RC1** release.

PR #116, opened 2017-11-08
DATACASS-509 - Request subsequent result pages in ReactiveSession asynchronously.

mp911de:
We now no longer require a `Scheduler` to offload `ResultSet`'s blocking paging but request the next result page asynchronously by adapting `ResultSet.fetchMoreResults()`. Result pages are requested once all elements of the previous `ResultSet` are emitted and the row publisher completes successfully. The Scheduler is no longer required.

The request progress is stored in a `MonoProcessor` to extend the result stream. Increase visibility of utility methods to avoid synthetic accessor creation.

---

Related ticket: [DATACASS-509](https://jira.spring.io/browse/DATACASS-509).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-cassandra-parent</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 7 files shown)


jxblum:
PR reviewed, polished an merged to {{master}} for SD Lovelace M1 and {{2.0.x}} for SD Kay RC2.

PR #118, opened 2018-01-19
DATACASS-512 - Support exists and count projections.

mp911de:
We now support exists and count projections derived from query methods and on Template API level. Exists queries fetch rows limiting to a single row to optimize fetching. Exists queries also support count results, i.e. a single row with a single numeric column is considered a count result. It's evaluated by comparing the number being greater to zero.

```java
boolean exists = template.exists(Query.query(where("firstname").is("Walter")), Person.class);

long count = template.count(Query.query(where("lastname").is("White")), Person.class);

interface PersonRepository extends Repository<Person, String> {

    long countByLastname(String lastname); // derived query

    boolean existsByLastname(String lastname);  // derived query

    @CountQuery("SELECT COUNT(*) from users WHERE lastname = ?0")
    long countQueryByLastname(String lastname); // String-based query

    @ExistsQuery("SELECT * from users WHERE lastname = ?0 LIMIT 1")
    boolean existsQueryByLastname(String lastname);  // String-based query
}
```

---

Related ticket: [DATACASS-512](https://jira.spring.io/browse/DATACASS-512).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-cassandra-parent</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
+	<version>2.1.0.DATACASS-512-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data for Apache Cassandra</name>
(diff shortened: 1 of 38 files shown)


jxblum:
Thank you for the PR; This PR has been reviewed, polished and merged to `master` in revision [c00e3f6](https://github.com/spring-projects/spring-data-cassandra/commit/c00e3f6b29c615de5ada2e959912ec9fad2eb600) for the upcoming **SD Lovelace M1** release.

PR #121, opened 2018-02-01
DATACASS-487 - Support map columns with mapped and converted key/value types.

mp911de:
We now support map columns that use mapped user-defined types and converted, non-primitive types in their keys and values. `Map` columns can be used for schema generation, column and key/value types are derived from the declared types by inspecting whether they are either UDTs or they can be converted by a custom registered converter.

```java
class Supplier {

  @Id String id;

  Map<Manufacturer, List<Currency>> currencies;
}

@UserDefinedType
class Manufacturer {
  String name;
}

class UDTToCurrencyConverter implements Converter<UDTValue, Currency> {
 // …
}

class CurrencyToUDTConverter implements Converter<Currency, UDTValue> {
  // …
}
```

---

Related ticket: [DATACASS-487](https://jira.spring.io/browse/DATACASS-487).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-cassandra-parent</artifactId>
-	<version>2.1.0.BUILD-SNAPSHOT</version>
+	<version>2.1.0.DATACASS-487-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data for Apache Cassandra</name>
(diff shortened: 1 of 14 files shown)


jxblum:
Reviewed, polished and merged to `master` for the _Spring Data (Cassandra)_ **Lovelace M1** release.

Agrawalmahin:
is this code merged?
if yes, can someone please the lowest version where this is available
I cannot see the changes in 3.0.9.RELEASE version

mp911de:
The change is merged since version 2.1. If you feel you found a bug and if you would like us to spend some time helping you to diagnose the problem, please spend some time describing it and, ideally, providing a [minimal sample](https://stackoverflow.com/help/mcve) that reproduces the problem in a new ticket.

PR #123, opened 2018-02-22
DATACASS-106 - Support persistence lifecycle callbacks

lukasz-antoniak:
JIRA ticket: https://jira.spring.io/browse/DATACASS-106

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
@@ -23,6 +23,11 @@ import java.util.stream.StreamSupport;
 import lombok.NonNull;
 import lombok.Value;
 
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.ApplicationEvent;
+import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.dao.DataAccessException;
 import org.springframework.data.cassandra.SessionFactory;
 import org.springframework.data.cassandra.core.convert.CassandraConverter;
@@ -41,6 +46,11 @@ import org.springframework.data.cassandra.core.cql.session.DefaultSessionFactory
 import org.springframework.data.cassandra.core.mapping.CassandraMappingContext;
 import org.springframework.data.cassandra.core.mapping.CassandraPersistentEntity;
 import org.springframework.data.cassandra.core.mapping.CassandraPersistentProperty;
+import org.springframework.data.cassandra.core.mapping.event.AfterDeleteEvent;
+import org.springframework.data.cassandra.core.mapping.event.AfterLoadEvent;
+import org.springframework.data.cassandra.core.mapping.event.AfterSaveEvent;
+import org.springframework.data.cassandra.core.mapping.event.BeforeDeleteEvent;
+import org.springframework.data.cassandra.core.mapping.event.BeforeSaveEvent;
 import org.springframework.data.cassandra.core.query.Query;
 import org.springframework.data.domain.Slice;
 import org.springframework.data.mapping.context.MappingContext;
@@ -81,7 +91,7 @@ import com.datastax.driver.core.querybuilder.Update;
  * @see org.springframework.data.cassandra.core.CassandraOperations
  * @since 2.0
  */
-public class CassandraTemplate implements CassandraOperations {
+public class CassandraTemplate implements CassandraOperations, ApplicationContextAware {
 
 	private final CassandraConverter converter;
 
@@ -92,6 +102,8 @@ public class CassandraTemplate implements CassandraOperations {
 	private final SpelAwareProxyProjectionFactory projectionFactory;
 
 	private final StatementFactory statementFactory;
+
+	private @Nullable ApplicationEventPublisher eventPublisher;
 
 	/**
 	 * Creates an instance of {@link CassandraTemplate} initialized with the given {@link Session} and a default
@@ -202,6 +214,15 @@ public class CassandraTemplate implements CassandraOperations {
 		return getMappingContext().getRequiredPersistentEntity(ClassUtils.getUserClass(entityType));
 	}
 
+	@org.springframework.lang.Nullable
+	private String guessTableName(Object entity) {
+		if (getMappingContext().hasPersistentEntityFor(entity.getClass())) {
+			return getMappingContext().getRequiredPersistentEntity(ClassUtils.getUserClass(entity.getClass())).getTableName().toCql();
+		}
+		// Not an entity.
+		return null;
+	}
+
 	/**
 	 * Returns a reference to the configured {@link ProjectionFactory} used by this template
 	 * to process CQL query projections.
@@ -290,7 +311,7 @@ public class CassandraTemplate implements CassandraOperations {
 		Assert.notNull(statement, "Statement must not be null");
 		Assert.notNull(entityClass, "Entity type must not be null");
 
-		Function<Row, T> mapper = getMapper(entityClass, entityClass);
+		Function<Row, T> mapper = getMapper(entityClass, entityClass, true);
 
 		return getCqlOperations().query(statement, (row, rowNum) -> mapper.apply(row));
 	}
@@ -306,7 +327,7 @@ public class CassandraTemplate implements CassandraOperations {
 
 		ResultSet resultSet = getCqlOperations().queryForResultSet(statement);
 
-		Function<Row, T> mapper = getMapper(entityClass, entityClass);
+		Function<Row, T> mapper = getMapper(entityClass, entityClass, true);
 
 		return QueryUtils.readSlice(resultSet, (row, rowNum) -> mapper.apply(row),
 				0, getEffectiveFetchSize(statement));
@@ -323,7 +344,7 @@ public class CassandraTemplate implements CassandraOperations {
 
 		ResultSet resultSet = getCqlOperations().queryForResultSet(statement);
 
-		return StreamSupport.stream(resultSet.spliterator(), false).map(getMapper(entityClass, entityClass));
+		return StreamSupport.stream(resultSet.spliterator(), false).map(getMapper(entityClass, entityClass, true));
 	}
 
 	/* (non-Javadoc)
@@ -352,7 +373,7 @@ public class CassandraTemplate implements CassandraOperations {
 
 	<T> List<T> doSelect(Query query, Class<?> entityClass, CqlIdentifier tableName, Class<T> returnType) {
 
-		Function<Row, T> mapper = getMapper(entityClass, returnType);
+		Function<Row, T> mapper = getMapper(entityClass, returnType, true);
 
 		RegularStatement select = getStatementFactory()
 				.select(query, getRequiredPersistentEntity(entityClass), tableName);
@@ -393,7 +414,7 @@ public class CassandraTemplate implements CassandraOperations {
 
 		ResultSet resultSet = getCqlOperations().queryForResultSet(statement);
 
-		return StreamSupport.stream(resultSet.spliterator(), false).map(getMapper(entityClass, returnType));
+		return StreamSupport.stream(resultSet.spliterator(), false).map(getMapper(entityClass, returnType, true));
 	}
 
 	/* (non-Javadoc)
@@ -578,8 +599,14 @@ public class CassandraTemplate implements CassandraOperations {
 
 		Insert insert = QueryUtils.createInsertQuery(tableName.toCql(), entity, options, getConverter());
 
+		maybeEmitEvent(new BeforeSaveEvent<Object>(entity, tableName.toCql(), insert));
+
 		// noinspection ConstantConditions
-		return getCqlOperations().execute(new StatementCallback(insert));
+		WriteResult result = getCqlOperations().execute(new StatementCallback(insert));
+
+		maybeEmitEvent(new AfterSaveEvent<Object>(entity, tableName.toCql()));
+
+		return result;
 	}
 
 	/* (non-Javadoc)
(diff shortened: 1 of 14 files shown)


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraOperations.java
```diff
@@ -312,7 +312,7 @@
 	 * @param entity The entity to insert, must not be {@literal null}.
 	 * @throws DataAccessException if there is any problem executing the query.
 	 */
-	void insert(Object entity) throws DataAccessException;
+	<T> void insert(T entity) throws DataAccessException;
```
mp911de: Let's keep the signatures as they were before.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
```diff
@@ -558,74 +579,92 @@ boolean doExists(Query query, Class<?> entityClass, CqlIdentifier tableName) {
...
 
 		Insert insert = QueryUtils.createInsertQuery(tableName.toCql(), entity, options, getConverter());
 
+		maybeEmitEvent(new BeforeSaveEvent<T>(entity, tableName.toCql()));
+
 		// noinspection ConstantConditions
-		return getCqlOperations().execute(new StatementCallback(insert));
+		WriteResult result = getCqlOperations().execute(new StatementCallback(insert));
+
+		maybeEmitEvent(new AfterSaveEvent<T>(entity, tableName.toCql()));
```
mp911de: Type args should not be required here.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
```diff
@@ -558,74 +579,92 @@ boolean doExists(Query query, Class<?> entityClass, CqlIdentifier tableName) {
...
 	@Override
-	public WriteResult update(Object entity, UpdateOptions options) {
+	public <T> WriteResult update(T entity, UpdateOptions options) {
 
 		Assert.notNull(entity, "Entity must not be null");
 		Assert.notNull(options, "UpdateOptions must not be null");
 
 		Update update = QueryUtils.createUpdateQuery(getTableName(entity).toCql(), entity, options, getConverter());
 
+		maybeEmitEvent(new BeforeSaveEvent<T>(entity, guessTableName(entity)));
```
mp911de: Any reason to not use above's `getTableName(entity)`?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
```diff
@@ -558,74 +579,92 @@ boolean doExists(Query query, Class<?> entityClass, CqlIdentifier tableName) {
...
+
 		// noinspection ConstantConditions
-		return getCqlOperations().execute(new StatementCallback(update));
+		WriteResult result = getCqlOperations().execute(new StatementCallback(update));
+
+		maybeEmitEvent(new AfterSaveEvent<T>(entity, guessTableName(entity)));
```
mp911de: We should reuse the table name

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
```diff
@@ -558,74 +579,92 @@ boolean doExists(Query query, Class<?> entityClass, CqlIdentifier tableName) {
...
 	@Override
-	public WriteResult delete(Object entity, QueryOptions options) {
+	public <T> WriteResult delete(T entity, QueryOptions options) {
 
 		Assert.notNull(entity, "Entity must not be null");
 		Assert.notNull(options, "QueryOptions must not be null");
 
 		Delete delete = QueryUtils.createDeleteQuery(getTableName(entity).toCql(), entity, options, getConverter());
 
+		maybeEmitEvent(new BeforeDeleteEvent<T>(entity, guessTableName(entity)));
```
mp911de: Any reason to not reuse `getTableName(entity)`?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
```diff
@@ -558,74 +579,92 @@ boolean doExists(Query query, Class<?> entityClass, CqlIdentifier tableName) {
...
+
 		// noinspection ConstantConditions
-		return getCqlOperations().execute(new StatementCallback(delete));
+		WriteResult result = getCqlOperations().execute(new StatementCallback(delete));
+
+		maybeEmitEvent(new AfterDeleteEvent<T>(entity, guessTableName(entity)));
```
mp911de: Any reason to not reuse `getTableName(entity)`?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
```diff
@@ -723,16 +762,21 @@ private int getEffectiveFetchSize(Statement statement) {
 	}
 
 	@SuppressWarnings("unchecked")
-	private <T> Function<Row, T> getMapper(Class<?> entityType, Class<T> targetType) {
+	private <T> Function<Row, T> getMapper(Class<?> entityType, Class<T> targetType, boolean emitEvents) {
```
mp911de: It would make sense to add the table name as an additional argument instead resolving it from the entity type. Is there any case we would not want to emit an event?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/event/AfterLoadEvent.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Event thrown after loading one or multiple rows which are further mapped into given type.
+ *
+ * @author Lukasz Antoniak
+ */
+public class AfterLoadEvent<E> extends CassandraMappingEvent<E> {
+	private static final long serialVersionUID = 1L;
+
+	public AfterLoadEvent(E source, @Nullable String table) {
```
mp911de: We should require the table name.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/event/AfterSaveEvent.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Event thrown after a single row has been inserted.
+ *
+ * @author Lukasz Antoniak
+ */
+public class AfterSaveEvent<E> extends CassandraMappingEvent<E> {
+	private static final long serialVersionUID = 1L;
+
+	public AfterSaveEvent(E source, @Nullable String table) {
```
mp911de: We should require the table name.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/event/BeforeDeleteEvent.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Event thrown before a single row is deleted.
+ *
+ * @author Lukasz Antoniak
+ */
+public class BeforeDeleteEvent<E> extends CassandraMappingEvent<E> {
+	private static final long serialVersionUID = 1L;
+
+	public BeforeDeleteEvent(E source, @Nullable String table) {
```
mp911de: We should require the table name.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/event/BeforeSaveEvent.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Event thrown before a single row has been inserted.
+ *
+ * @author Lukasz Antoniak
+ */
+public class BeforeSaveEvent<E> extends CassandraMappingEvent<E> {
+	private static final long serialVersionUID = 1L;
+
+	public BeforeSaveEvent(E source, @Nullable String table) {
```
mp911de: We should require the table name.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/event/CassandraMappingEvent.java
```diff
@@ -0,0 +1,58 @@
...
+	private static final long serialVersionUID = 1L;
+	private final @Nullable String table;
+
+	/**
+	 * Creates new {@link CassandraMappingEvent}.
+	 *
+	 * @param source must not be {@literal null}.
+	 * @param table may be {@literal null}.
+	 */
+	public CassandraMappingEvent(T source, @Nullable String table) {
```
mp911de: We should require the table name.

src/main/asciidoc/reference/mapping.adoc
```diff
@@ -582,3 +582,33 @@ Below is an example of a Spring `Converter` implementation that converts from a
...
+  }
+}
+----
+====
+
+Simply declaring these beans in your Spring `ApplicationContext` will cause them to be invoked whenever the event is dispatched.
+
+The list of callback methods that are present in AbstractMappingEventListener are:
+
+* `onBeforeSave` - called in `CassandraTemplate#insert(...)` operation *before* inserting/updating record in the database.
```
mp911de: You might want to include `update(…)` as well.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
```diff
@@ -202,6 +214,15 @@ public CqlOperations getCqlOperations() {
 		return getMappingContext().getRequiredPersistentEntity(ClassUtils.getUserClass(entityType));
 	}
 
+	@org.springframework.lang.Nullable
+	private String guessTableName(Object entity) {
```
mp911de: How about reusing `getTableName(…)`?

mp911de:
Ping! Any update on this one? The pull request looks very promising and we'd like to incorporate your contribution to our codebase.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
lukasz-antoniak: Currently there is no use-case for not emitting events, but I felt that this should be configurable. In the end, caller would like to receive a mapper and may not care about triggering events. Matter of taste.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
lukasz-antoniak: At first I have reused `#getTableName(...)`. The only issue is that `CustomConversionIntegrationTests` fails. `#getTableName(...)` raises exception when provided Java type is not an entity. That is the reason of introducing the `#guessTableName(...)` method and making table name optional in the event POJOs.

lukasz-antoniak:
Apologies for delay. I have added support for reactive template and included CQL `Statement` in before insert/update/delete events.

mp911de:
All good, no worries, four days isn't a real delay, more an extended weekend 🙃

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
mp911de: When `CustomConversionIntegrationTests` starts failing, then we need to investigate the actual cause and how to fix it. Happy to support you there.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
mp911de: We can add that kind of things always at a later stage. This method is `private` and nothing breaks if we change that method at a later time.

matthewadams:
Hey, I'm glad to see this issue getting some love.  Nice work, @lukasz-antoniak!

While this issue & PR are specific to Spring Data Cassandra, I would remind everyone that this is more of a generic feature that, IMHO, deserves to be in Spring Data Commons.  There's nothing Cassandra-specific about instance callbacks.  Heck, they go all the way back to [JDO](https://db.apache.org/jdo/api30/apidocs/javax/jdo/InstanceCallbacks.html).  I'd be careful not to introduce Cassandra-specifics into the APIs so that they could be shared amonst all Spring Data subprojects.

Thoughts?

lukasz-antoniak:
@mp911de feel free to polish it from here :). I will definitely have a look once you push it to master.

mp911de:
@matthewadams We considered lifecycle events on Commons level a couple times but ended up not adding these as we want to allow extensibility and hooks for customization of store-specific objects. We want to allow e.g. to change the domain object or adjust the to-be-executed statement before saving/after loading which pulls in store-specific API.

Some stores bring their own loading mechanisms and eventing (JPA, Neo4j) where listeners do not align to reflect a consistent behavior.

matthewadams:
> Some stores bring their own loading mechanisms and eventing (JPA, Neo4j) where listeners do not align to reflect a consistent behavior.

Yeah, @mp911de, I forgot about that.  Good point.

mp911de:
@lukasz-antoniak I polished your contribution and pushed it to our issue branch [issue/DATACASS-106](https://github.com/spring-projects/spring-data-cassandra/compare/issue/DATACASS-106?expand=1). Care to have a look before I push the changes to `master`?

lukasz-antoniak:
@mp911de looks great. Next time I will try to submit code requiring less work from your side :).

mp911de:
That's merged and polished now. Thanks a lot for your contribution.

PR #128, opened 2018-05-14
DATACASS-529 - Allow slice queries using reactive repositories.

hleb-albau:
Reactive repositories now support methods returning `Mono<Slice<T>>`.
```
Mono<Slice<User>> findByLastname(String firstname, Pageable pageable);
```

Related ticket: DATACASS-529.

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/QueryUtils.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/QueryUtils.java
@@ -17,7 +17,9 @@ package org.springframework.data.cassandra.core;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Function;
 
+import org.springframework.data.cassandra.ReactiveResultSet;
 import org.springframework.data.cassandra.core.cql.QueryOptions;
 import org.springframework.data.cassandra.core.cql.QueryOptionsUtil;
 import org.springframework.data.cassandra.core.cql.RowMapper;
@@ -29,9 +31,11 @@ import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Slice;
 import org.springframework.data.domain.SliceImpl;
 import org.springframework.util.Assert;
+import reactor.core.publisher.Mono;
 
 import com.datastax.driver.core.PagingState;
 import com.datastax.driver.core.ResultSet;
+import com.datastax.driver.core.Row;
 import com.datastax.driver.core.querybuilder.Delete;
 import com.datastax.driver.core.querybuilder.Delete.Where;
 import com.datastax.driver.core.querybuilder.Insert;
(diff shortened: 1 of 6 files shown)


mp911de:
Thanks a lot, we'll take a look as soon as we have time.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/QueryUtils.java
```diff
@@ -165,4 +169,23 @@ static Delete createDeleteQuery(String tableName, Object objectToDelete, QueryOp
...
+
+	/**
+	 * Read a {@link Mono<Slice>} of data from the {@link ReactiveResultSet} for a {@link Pageable}.
+	 *
+	 * @param resultSet must not be {@literal null}.
+	 * @param fetchSize must not be {@literal null}.
+	 * @param rowMapper must not be {@literal null}.
+	 * @return the resulting {@link Slice}.
+	 */
+	static <T> Mono<Slice<T>> readSlice(ReactiveResultSet resultSet, Integer fetchSize, Function<Row, T> rowMapper) {
```
mp911de: We can't introduce dependencies to reactive types here as we want the module to run without reactive dependencies as well.

We also need to be careful about fetching/reading progress. If the fetch size differs from the actual number of rows returned for a page, we might request the next page without intending to do so. I think we need another method on `ReactiveResultSet` that gives us a `Flux` without transparent next page fetching.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/ReactiveCassandraOperations.java
```diff
@@ -97,6 +99,19 @@
...
 	<T> Flux<T> select(Statement statement, Class<T> entityClass) throws DataAccessException;
 
+	/**
+	 * Execute a {@code SELECT} query with paging and convert the result set to a {@link Slice} of entities.
+	 *
+	 * A sliced query translates the effective {@link Statement#getFetchSize() fetch size} to the page size.
+	 *
+	 * @param statement the CQL statement, must not be {@literal null}.
+	 * @param entityClass The entity type must not be {@literal null}.
+	 * @return the result object returned by the action or {@link Mono#empty()}
```
mp911de: In this case, we're mixing paradigms. We should return an empty `Slice` (see `Page.empty()`) as the query terminates without rows and the query result is the container for rows, not the rows themselves.

hleb-albau:
Hi @mp911de,
I have no free time in near future, so if you can finish this task, it would be nice. Thanks

mp911de:
Cool, thanks for your timely update. We'll take care of this one.

eximius313:
@mp911de do you have any plans when you could handle this?

mp911de:
This pull request needs some polishing before we can merge it. We intend to ship the change with the next release candidate.

eximius313:
So `2.1.0.RC1`?

mp911de:
@lukasz-gosiewski Please file a new ticket in [Spring Data Cassandra](https://jira.spring.io/browse/DATACASS). We'll see what we can do for you.

mp911de:
That's merged and polished now. Thanks a lot!

PR #132, opened 2018-05-24
Full editing pass for spring-data-cassandra

Buzzardo:
I edited for spelling, punctuation, grammar, usage, and corporate voice. I also added an epub cover image and the SVG file from which it is generated.


mp911de:
Thanks a lot. That's merged and polished now.

PR #133, opened 2018-06-04
DATACASS-555 - minor fix to CqlTemplate

CodeSmell:
Allow for caching the prepared statements in the CqlTemplate if a PreparedStatementCache implementation is provided
https://jira.spring.io/browse/DATACASS-555

- [X] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [X] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATACASS).
- [X] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [X] You submit test cases (unit or integration tests) that back your changes.
- [X] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/CqlTemplate.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/CqlTemplate.java
@@ -69,6 +69,8 @@ import com.datastax.driver.core.exceptions.DriverException;
  * @author Antoine Toulme
  * @author John Blum
  * @author Mark Paluch
+ * @author Mike Barlotta (CodeSmell)
+ *
  * @see PreparedStatementCreator
  * @see PreparedStatementBinder
  * @see PreparedStatementCallback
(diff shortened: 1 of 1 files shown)


CodeSmell:
@mp911de 

Created separate PR for this when saw the build failing.
Running `mvn clean test` locally works, but that is b/c my Cassandra version is `3.11.2`. 

The local time tests, when Cassandra version is 2.2.10, are failing in `CassandraTypeMappingIntegrationTests`

A quick look and can't see why these changes would impact this 
`operations.insert(entity);`

Seems related to this
https://gist.github.com/mp911de/b61a1a24bd824d0742a6b2e31ecb4861

As the error is 
```nested exception is com.datastax.driver.core.exceptions.InvalidQueryException: Invalid INTEGER constant (3723000) for "localtime" of type time
	at ```

an thoughts?

mp911de:
This issue was fixed in Cassandra 3.0.8 and 3.8 with [CASSANDRA-11798](https://issues.apache.org/jira/browse/CASSANDRA-11798). We require symmetric data types and the problem with `time` was that reading-wise it returned a number whereas updating required a string literal. `time` is only supported on Cassandra versions 3.0.8/3.8 and higher.

See also:
* https://jira.spring.io/browse/DATACASS-302
* https://issues.apache.org/jira/browse/CASSANDRA-11798

mp911de:
I filed [DATACASS-563](https://jira.spring.io/browse/DATACASS-563) to address the failing tests with Cassandra 2.2.

mp911de:
Thanks a lot. That's merged, polished, and backported now.

PR #134, opened 2018-07-19
Support for reactive batch ops 

OlegDokuka:
https://jira.spring.io/browse/DATACASS-574

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/ReactiveCassandraBatchOperations.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/ReactiveCassandraBatchOperations.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright 2016-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


OlegDokuka:
@mp911de ready to review. 

Do we really need support for inserting `Publisher`? From my perspective, we will be able to execute elements draining only on the `.execute` method invocation. In turn, it will not give any benefits since we will have to collect data before writing, so in any case, it will be kind of "waiting point" + there is danger point that one sent infinitive stream, etc

mp911de:
Thanks for raising this item. Accepting `Publisher` is an optimization which brings a challenge of batching and awaiting. In MongoDB, we switched to accept `Mono<List<T>>` to allow deferred inserts. Using the API would require to collect items upfront (`.buffer()` or `.collectList()`) and pass in the collected objects for a batch.

For the start, we're good without accepting `Publisher`.

OlegDokuka:
I will add support for Publisher. Let me check Mongo api first in order to be aligned.

OlegDokuka:
@mp911de done

mp911de:
Thanks a lot. We're heading right now towards 2.1 **RC1** so this pull request is a candidate for the second release candidate, 2.1 **RC2**.

OlegDokuka:
@mp911de Great! Do I need to add/change something there or it is good to good but will be merged later?

mp911de:
I haven't reviewed the pull request yet as we're currently busy with preparing 2.1 RC1.

mp911de:
Thanks for your contribution. I had to convert spaces to tabs (please apply [our formatting settings next time](https://github.com/spring-projects/spring-data-build/blob/master/etc/ide/eclipse-formatting.xml)) and applied a round of polishing. You might be interested in https://github.com/spring-projects/spring-data-cassandra/commit/8055890ac5326881d21dd0bab444e84c54713479.

OlegDokuka:
hey @mp911de. Thank you for polishing. Should I add something else there

mp911de:
Nothing left to do here.

PR #135, opened 2019-01-29
DATACASS-575 - Support IF conditions in lightweight transactions with UPDATE and DELETE

mp911de:
We now support conditions in lightweight transactions for `UPDATE` and `DELETE` statements. Conditions are `Filter` objects similar to the `WHERE` clause. Conditions are supported for entity and query-based update/delete methods.

```java
UpdateOptions options = UpdateOptions.builder().ifCondition(where("firstname").is("Walter")).build();
User user = new User("heisenberg", "Walter", "White");
template.update(user, options);

DeleteOptions options = DeleteOptions.builder().ifCondition(where("counter").is(42)).build();
Query query = Query.query(where("id").is("heisenberg")).queryOptions(options);
template.delete(query, User.class);
```

---

Related ticket: [DATACASS-575](https://jira.spring.io/browse/DATACASS-575).

PR #136, opened 2019-01-30
DATACASS-576 - Add support for Optimistic Locking.

mp911de:
We now support Optimistic Locking for insert, update and delete operations leveraging Cassandra's lightweight transaction support. Modifying statements are enhanced with `IF` conditions to conditionally insert and modify rows and to prevent concurrent modifications by throwing `OptimisticLockingFailureException`.

```java
@Table
class Person {

  @Id String id;
  String firstname;
  String lastname;
  @Version Long version;
}

Person daenerys = template.insert(new Person("Daenerys"));

Person tmp = template.findOne(query(where("id").is(daenerys.getId())), Person.class);

daenerys.setLastname("Targaryen");
template.save(daenerys);

template.save(tmp); // throws OptimisticLockingFailureException
```

---

Depends on: [DATACASS-575](https://jira.spring.io/browse/DATACASS-575), #135.
Related ticket: [DATACASS-576](https://jira.spring.io/browse/DATACASS-576).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-cassandra-parent</artifactId>
-	<version>2.2.0.BUILD-SNAPSHOT</version>
+	<version>2.2.0.DATACASS-575-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data for Apache Cassandra</name>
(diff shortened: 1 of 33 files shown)


jxblum:
We may want to consider support for version properties other than of type `long`, or primitive types in general.

Often times, it is even useful to use multiple properties in optimistic locking (e.g. user + timestamp).

Our initial implementation should satisfy most UCs for the time being.

mp911de:
@jxblum care to file a ticket against Spring Data Commons to discuss the idea?

PR #137, opened 2019-02-06
DATACASS-599 Restore compatibility with Guava 26+

MateuszStefek:
Guava has removed Futures#transform method without the Executor parameter in version 26.0-jre.
DefaultBridgeReactiveSession fails with NoSuchMethod exception when used with Guava 26.0-jre+.


mp911de:
Thanks a lot. That's merged, polished and backported now.

PR #158, opened 2019-04-03
DATACASS-647 - Add non-nullable variant to ReactiveSelectOperation extensions.

sdeleuze:


mp911de:
That's merged now. Thanks a lot.

PR #166, opened 2019-10-24
DATACASS-694 - Fix LocalTime conversion

hurelhuyag:
java.time.LocalTime to Cassandra time datatype mapping fix

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/convert/CassandraJsr310Converters.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/convert/CassandraJsr310Converters.java
@@ -106,7 +106,7 @@ public abstract class CassandraJsr310Converters {
 
 		@Override
 		public LocalTime convert(Long source) {
-			return LocalTime.ofNanoOfDay(TimeUnit.MILLISECONDS.toNanos(source));
+			return LocalTime.ofNanoOfDay(source);
 		}
 	}
 
@@ -124,7 +124,7 @@ public abstract class CassandraJsr310Converters {
 
 		@Override
 		public Long convert(LocalTime source) {
-			return source.getLong(ChronoField.MILLI_OF_DAY);
+			return source.getLong(ChronoField.NANO_OF_DAY);
 		}
 	}
 }
--- a/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/core/convert/CassandraJsr310ConvertersUnitTests.java
+++ b/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/core/convert/CassandraJsr310ConvertersUnitTests.java
@@ -34,7 +34,7 @@ public class CassandraJsr310ConvertersUnitTests {
 	@Test // DATACASS-302
 	public void shouldConvertLongToLocalTime() {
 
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@hurelhuyag Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-cassandra&pullRequestId=166)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-cassandra&pullRequestId=166) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@hurelhuyag Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-cassandra&pullRequestId=166)!

mp911de:
Care to add an integration test to `org.springframework.data.cassandra.core.convert.CassandraTypeMappingIntegrationTests` that asserts the Cassandra-specific representation of  `time`?

hurelhuyag:
Read and Write test will work fine. Not needed to add more. Only way to check that is use cqlsh

mp911de:
We have existing tests that prove that read and write of `LocalTime` works as expected. Applying a change to the conversion logic does not break existing tests and this is something to worry about. The reason tests didn't break is because conversion follows the same contract and our tests verify values after conversion.

Previously, the representation in Cassandra (`Row`) didn't meet how Cassandra represents `time`, therefore it is indeed necessary to add further tests to verify the Cassandra representation as seen through the driver.

hurelhuyag:
On the other hand datastax documented time datatype is nanoseconds not milliseconds.

> Value is encoded as a 64-bit signed integer representing the number of nanoseconds since midnight. Values can be represented as strings, such as 13:30:54.234.

https://docs.datastax.com/en/archived/cql/3.3/cql/cql_reference/cql_data_types_c.html#cql_data_types_c__cqlDataTypes

hurelhuyag:
And Cassandra `Row.getTime()` method returns long in nanoseconds

> long getTime(int i)
Returns the ith value as a long in nanoseconds since midnight.
This method uses the CodecRegistry to find a codec to convert the underlying CQL type to a Java long (for CQL type time, this will be the built-in codec).

hurelhuyag:
And I'm not sure this kind of test is needed or not

```
@Test
	public void shouldTest(){
		assumeTrue(cassandraVersion.isGreaterThanOrEqualTo(VERSION_3_10));
		AllPossibleTypes entity = new AllPossibleTypes("1");

		entity.setLocalTime(java.time.LocalTime.of(1, 2, 3));

		operations.insert(entity);

		ResultSet resultSet = session.execute("SELECT localTime FROM AllPossibleTypes WHERE id = '1'");
		Iterator<Row> rowIt = resultSet.iterator();
		while (rowIt.hasNext()){
			Row row = rowIt.next();
			long timeNanos = row.getTime(0);
			assertThat(timeNanos).isEqualTo(3723000000000L);
		}
	}
```

mp911de:
@hurelhuyag it is needed to prevent that this type of bug comes again.

hurelhuyag:
2 integration test added

mp911de:
Thanks a lot for your contribution. That's squashed, merged, polished, and backported now.

PR #172, opened 2020-03-31
DATACASS-465 - Introduces `@Frozen` annotation to mark schema elements as Frozen.

schauder:
`@Frozen` is supported on UDTs, collection like properties (`Set`, `List` and `Map`), and their type parameters if those are UDTs.

For example

    @Frozen Map<String, @Frozen MyUdt> propertyName;

Issue: https://jira.spring.io/browse/DATACASS-465


mp911de:

PR #173, opened 2020-04-03
DATACASS-167 - Add support for embedded entities.

christophstrobl:
Embedded entities are used to have value objects in your java data model who's properties are flattened out into the table.
In the following example you see, that `User.name` is annotated with `@Embedded`.
The consequence of this is the properties of `UserName` are folded into the `user` table which consists of 3 columns (`user_id`, `firstname`, `lastname`).

Embedded entities may only contain simple property types. It is not possible to nest an embedded entity into another embedded one.

However, if the `firstname` and `lastname` column values are actually `null` within the result set, the entire property `name` will be set to `null` according to the `onEmpty` of `@Embedded`, which ``null``s objects when all nested properties are `null`.

Opposite to this behaviour `USE_EMPTY` tries to create a new instance using either a default constructor or one that accepts nullable parameter values from the result set.

```java
public class User {

    @PrimaryKey("user_id")
    private String userId;

    @Embedded(onEmpty = USE_NULL)
    UserName name;
}

public class UserName {
    private String firstname;
    private String lastname;
}
```

If you need a value object multiple times in an entity, this can be achieved with the optional `prefix` element of the `@Embedded` annotation.
This element represents a prefix and is prepend for each column name in the embedded object.

The `@Indexed` annotation can be applied to single properties of embedded entities or along side with the `@Embedded` annotation, in which case all properties of the embedded are indexed.

Query derivation follows the general rules using either camel case or explicit underscore notation when accessing properties of the embedded type in a repository query method.

```java
public interface UserRepository extends CassandraRepository<User, String> {
    User findByName_Firstname(String firstname);
}
``` 

On template level paths to embedded properties use the `.` (dot) notation.

```java
Criteria.where("name.firstname").is("walter")
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
 	<groupId>org.springframework.data</groupId>
 	<artifactId>spring-data-cassandra-parent</artifactId>
-	<version>3.0.0.BUILD-SNAPSHOT</version>
+	<version>3.0.0.DATACASS-167-SNAPSHOT</version>
 	<packaging>pom</packaging>
 
 	<name>Spring Data for Apache Cassandra</name>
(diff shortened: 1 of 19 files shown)


mp911de:

PR #177, opened 2020-08-06
DATACASS-767: Allow configuring keyspace per Statement

tomekl007:

### DIFF
--- a/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/util/CassandraDelegate.java
+++ b/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/util/CassandraDelegate.java
@@ -28,12 +28,14 @@ import org.springframework.data.cassandra.support.CqlDataSet;
 import org.springframework.util.Assert;
 import org.springframework.util.SocketUtils;
 import org.springframework.util.StringUtils;
-
 import org.testcontainers.containers.CassandraContainer;
 
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.CqlSessionBuilder;
+import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
+import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
+import com.datastax.oss.driver.api.core.metadata.Node;
 
 /**
  * Delegate used to provide a Cassandra context for integration tests. This rule can use/spin up either an embedded
@@ -43,6 +45,7 @@ import com.datastax.oss.driver.api.core.CqlSessionBuilder;
  *
  * @author Mark Paluch
  * @author John Blum
+ * @author Tomasz Lelek
  * @since 1.5
  * @see CassandraConnectionProperties
  */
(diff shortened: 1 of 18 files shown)


pivotal-issuemaster:
@tomekl007 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-cassandra&pullRequestId=177)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-cassandra&pullRequestId=177) to manually synchronize the status of this Pull Request.

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/util/CassandraDelegate.java
```diff
@@ -297,7 +299,8 @@ private CqlSessionBuilder buildCluster(int port) {
 		String host = resolveHost();
 
 		return CqlSession.builder().addContactPoint(InetSocketAddress.createUnresolved(host, port))
-				.withLocalDatacenter("datacenter1");
+				.withLocalDatacenter("datacenter1").withConfigLoader(
```
tomekl007, line 302: The new `setKeyspace` per statement behavior is working from Cassandra `4.0` and `V5` protocol version:
1. It seems that the existing integration tests do not support passing the parameter from the test to the session creation logic. What is the best way to parameterize tests(`ReactiveCqlTemplateIntegrationTests.java`, `CqlTemplateIntegrationTests.java`, `CassandraTemplateIntegrationTests.java) that are using `V5` and construct specialized session for them?
2. The `V5` is supported only in Cassandra 4.0. Your test infrastructure is taking docker images from here: https://hub.docker.com/_/cassandra, where there is no 4.0 yet. Do you have any other way to run ITs for Cassandra 4.0, or should we submit the PR with Cassandra 4.0 to https://github.com/docker-library/cassandra?
3. If you upgrade our java-driver to `4.8.x`, the `V5` tests will not work correctly with Cassandra 4.0. The reason is that java-driver `4.8.x` for `V5` is using a new checksumming (https://github.com/datastax/java-driver/pull/1458) that is not yet supported at the Cassandra 4.x server-side.

pivotal-issuemaster:
@tomekl007 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-cassandra&pullRequestId=177)!

spring-data-cassandra/src/test/java/org/springframework/data/cassandra/test/util/CassandraDelegate.java
mp911de: Integration tests can be run against:

* A local Cassandra instance
* An embedded one
* via TestContainers

I'd suggest performing a version check against the configured Cassandra instance. If it is version 4+, then we default to protocol `V5`. Otherwise, we stick with driver defaults until we can safely assume that the driver can perform a handshake itself using the newest version and fall back to the highest supported version.

tomekl007: So basically, what you are proposing is something like this:
```
		String cassandra_version = System.getenv("CASSANDRA_VERSION");

		if (Version.parse(cassandra_version).isGreaterThanOrEqualTo(Version.parse("4.0"))) {
			return CqlSession.builder().addContactPoint(InetSocketAddress.createUnresolved(host, port))
					.withLocalDatacenter("datacenter1").withConfigLoader(
							DriverConfigLoader.programmaticBuilder().withString(DefaultDriverOption.PROTOCOL_VERSION, "V5").build());

		} else {
			return CqlSession.builder().addContactPoint(InetSocketAddress.createUnresolved(host, port))
					.withLocalDatacenter("datacenter1");
		}
```
right? It will work properly, but only for test-containers. 

I don't see a clear way (in `CassandraDelegate`) to detect the Cassandra version before creating the session when the ITs are running in embedded or local Cassandra mode.

mp911de: Something along those lines. I was actually suggesting to create a `CqlSession` without altering the protocol version first to obtain the Cassandra version. Then close the `CqlSession` and based on the version, then create the actual `CqlSession` that is used for tests.

tomekl007: Implemented in:
https://github.com/spring-projects/spring-data-cassandra/pull/177/commits/07f06157ffdbb5e91c02b00090a2debeab0ac7dd
let me know if this is what you mean.

tomekl007:
> Amazing work, thank you for your contribution.
> 
> There are a few minor stylistic issues that we can sort out during the merge. Can you please squash all your changes into a single commit as we typically go with a single commit plus a polishing commit for a ticket?

All commits squashed into a single one.

mp911de:
Thank you for your contribution. That's merged and polished now. Mostly reformatting (alphabetically ordering of `QueryOptions`) and enabling keyspace propagation for prepared statements.

Ślicznie dziękuję.

PR #178, opened 2020-08-10
DATACASS-766 Add support for port per endpoint

tomekl007:

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
@@ -23,6 +23,9 @@ import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.function.IntFunction;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.slf4j.Logger;
@@ -64,6 +67,7 @@ import com.datastax.oss.driver.api.core.CqlSessionBuilder;
  * @author Matthew T. Adams
  * @author John Blum
  * @author Mark Paluch
+ * @author Tomasz Lelek
  * @since 3.0
  */
 public class CqlSessionFactoryBean
@@ -107,12 +111,22 @@ public class CqlSessionFactoryBean
 
 	private @Nullable SessionBuilderConfigurer sessionBuilderConfigurer;
 
-	private String contactPoints = DEFAULT_CONTACT_POINTS;
+	private IntFunction<Collection<InetSocketAddress>> contactPoints = port -> createInetSocketAddresses(
+			DEFAULT_CONTACT_POINTS, port);
 
 	private @Nullable String keyspaceName;
 	private @Nullable String localDatacenter;
 	private @Nullable String password;
 	private @Nullable String username;
+	private Supplier<CqlSessionBuilder> setCqlSessionBuilderSupplier;
+
+	public CqlSessionFactoryBean() {
+		this(CqlSession::builder);
+	}
+
+	CqlSessionFactoryBean(Supplier<CqlSessionBuilder> setCqlSessionBuilderSupplier) {
+		this.setCqlSessionBuilderSupplier = setCqlSessionBuilderSupplier;
+	}
 
 	/**
 	 * Null-safe operation to determine whether the Cassandra {@link CqlSession} is connected or not.
@@ -130,13 +144,24 @@ public class CqlSessionFactoryBean
 
 	/**
 	 * Set a comma-delimited string of the contact points (hosts) to connect to. Default is {@code localhost}; see
-	 * {@link #DEFAULT_CONTACT_POINTS}.
+	 * {@link #DEFAULT_CONTACT_POINTS}. It can be in the form 'host:port', or a simple 'host' to use the configured port.
 	 *
 	 * @param contactPoints the contact points used by the new cluster.
 	 */
 	public void setContactPoints(String contactPoints) {
-		this.contactPoints = contactPoints;
+		this.contactPoints = port -> createInetSocketAddresses(contactPoints, port);
 	}
+
+	/**
+	 * Set a collection of the contact points (hosts) to connect to. Default is {@code localhost}; see
+	 * {@link #DEFAULT_CONTACT_POINTS}.
+	 *
+	 * @param contactPoints the contact points used by the new cluster.
+	 */
+	public void setContactPoints(Collection<InetSocketAddress> contactPoints) {
+		this.contactPoints = unusedPort -> contactPoints;
+	}
+
 
 	/**
 	 * Sets the name of the local datacenter.
@@ -439,13 +464,12 @@ public class CqlSessionFactoryBean
 	}
 
 	protected CqlSessionBuilder buildBuilder() {
+		Collection<InetSocketAddress> addresses = contactPoints.apply(this.port);
+		Assert.notEmpty(addresses, "At least one server is required");
 
-		Assert.hasText(this.contactPoints, "At least one server is required");
+		CqlSessionBuilder sessionBuilder = setCqlSessionBuilderSupplier.get();
 
-		CqlSessionBuilder sessionBuilder = CqlSession.builder();
-
-		StringUtils.commaDelimitedListToSet(this.contactPoints).forEach(host ->
-			sessionBuilder.addContactPoint(InetSocketAddress.createUnresolved(host, this.port)));
+		addresses.forEach(sessionBuilder::addContactPoint);
 
 		if (StringUtils.hasText(this.username)) {
 			sessionBuilder.withAuthCredentials(this.username, this.password);
@@ -458,6 +482,45 @@ public class CqlSessionFactoryBean
 		return this.sessionBuilderConfigurer != null
 			? this.sessionBuilderConfigurer.configure(sessionBuilder)
 			: sessionBuilder;
+	}
+
+	private Collection<InetSocketAddress> createInetSocketAddresses(String contactPoints, int port) {
+		return StringUtils.commaDelimitedListToSet(contactPoints).stream().map(candidate -> toHostAndPort(candidate, port))
+				.map(hostAndPort -> InetSocketAddress.createUnresolved(hostAndPort.host, hostAndPort.port))
+				.collect(Collectors.toList());
+	}
+
+	private static class HostAndPort {
+		private final String host;
+		private final int port;
+
+		private HostAndPort(String host, int port) {
+			this.host = host;
+			this.port = port;
+		}
+	}
+
+	private HostAndPort toHostAndPort(String candidate, int port) {
+		int i = candidate.lastIndexOf(':');
+		if (i == -1 || !isPort(() -> candidate.substring(i + 1))) {
+			return new HostAndPort(candidate, port);
+		} else {
+			String[] hostAndPort = candidate.split(":");
+			if (hostAndPort.length != 2) {
+				throw new IllegalArgumentException(
+						String.format("The provided contact point: %s has wrong format.", candidate));
+			}
+			return new HostAndPort(hostAndPort[0], Integer.parseInt(hostAndPort[1]));
+		}
+	}
+
+	private boolean isPort(Supplier<String> value) {
+		try {
+			int i = Integer.parseInt(value.get());
+			return i > 0 && i < 65535;
+		} catch (Exception ex) {
+			return false;
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
```diff
@@ -138,6 +152,19 @@ public void setContactPoints(String contactPoints) {
...
+	/**
+	 * Set a comma-delimited string of the contact points (hosts) with port to connect to. Default is null, meaning that
+	 * this method will be ignored and standard {@link CqlSessionFactoryBean#setContactPoints(String)} and
+	 * {@link CqlSessionFactoryBean#setPort(int)} will apply. If set, it will have priority over the aforementioned
+	 * methods.
+	 *
+	 * @param contactPointsWithPort the contact points with port used by the new cluster. The format of this setting
+	 *          should be following: host:port,host2:port2
+	 */
+	public void setContactPointsWithPort(String contactPointsWithPort) {
```
tomekl007, line 164: I've added a new method that works in a backward-compatible way:
- if the old solution: `addContactPoints` and `addPort` is used it works the same way
- if someone intends to specify the port per contact point, the new API can be used and it has priority over the old API.
Also, I am not sure about the type of the argument. Maybe it will be better to accept here a list of addresses? 
Or even list of tuples (or something similar) with `String host` and `int port`?
Right now I've created one string argument to have this API unified with the existing `setContactPoints` but because of the fact that `String` is used, there is validation required (see my 2nd comment).
wdyt?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
```diff
@@ -113,6 +117,16 @@
...
 	private @Nullable String password;
 	private @Nullable String username;
+	private Supplier<CqlSessionBuilder> setCqlSessionBuilderSupplier;
+	private @Nullable String contactPointsWithPort;
+
+	public CqlSessionFactoryBean() {
+		this(CqlSession::builder);
+	}
+
+	CqlSessionFactoryBean(Supplier<CqlSessionBuilder> setCqlSessionBuilderSupplier) {
```
tomekl007, line 127: this package-private constructor is created for unit testing purpose (mocking)

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
```diff
@@ -442,10 +469,22 @@ protected CqlSessionBuilder buildBuilder() {
...
-			sessionBuilder.addContactPoint(InetSocketAddress.createUnresolved(host, this.port)));
+		CqlSessionBuilder sessionBuilder = setCqlSessionBuilderSupplier.get();
+
+		if (contactPointsWithPort == null) {
+			StringUtils.commaDelimitedListToSet(this.contactPoints)
+					.forEach(host -> sessionBuilder.addContactPoint(InetSocketAddress.createUnresolved(host, this.port)));
+		} else {
+			StringUtils.commaDelimitedListToSet(this.contactPointsWithPort).stream().map(HOST_PORT_SPLITTER::splitToList)
+					.forEach(hostAndPort -> {
+						if (hostAndPort.size() != 2) {
```
tomekl007, line 480: If we decide to keep one `String` argument, then I think that we should do some validation. Right now I am doing a very simple check - if any `hostAndPort` is corrupted, the `IllegalArgumentException` with ALL `String contactPointsWithPort` is thrown. 
I can rework it to be more detailed, that is iterate over every pair of host and port, collect info about the corrupted ones, and report only those in the `IllegalArgumentException` message. Let me know wdyt about such detailed information - maybe the current validation is good enough?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
```diff
@@ -55,6 +56,7 @@
 
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.CqlSessionBuilder;
+import com.datastax.oss.driver.shaded.guava.common.base.Splitter;
```
mp911de, line 59: Let's stick with core Java or Spring API (e.g. `org.springframework.util.StringUtils.split(…)`).

tomekl007:
> There's nothing wrong with slightly extending the definition of `contactPoints` that they may contain a port per contact point. Since we're splitting and parsing `contactPoints` anyway, we should leverage the existing property instead of introducing a new one.

what about backward compatibility in that case? If we will alter the behavior of the existing `setContactPoints` method, client code that is using `setContactPoints` and `setPort` will break. 
Also, we won't be able to know if the port from `setPort` property should be taken, or should we use the port provided in the `contactPoints` string.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
tomekl007: The `StringUtils.split` was my first choice, but this method is returning only two elements: before and after the delimiter. So when using `StringUtils.split`, we won't be able to detect such a situation:
https://github.com/spring-projects/spring-data-cassandra/pull/178/files#diff-dbb3a2d9e16f999f5196b32b5e9820faR90

mp911de: We can also default to `String.split(',')` which should do the same job. Since we don't expect a comma in the hostname or an IPv6 address, we should be safe unless I'm missing a subtlety.

mp911de:
Backward compatibility is given as `contactPoints` expects one or more hosts without a port. If we discover a hostname without a port, then we use the `port` value that was set on the factory bean. If a contact point was entered with a port, then the port of the contact point has precedence over `setPort(…)`. It makes sense to reflect this behavior in the Javadocs of both setters.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
mp911de: Argument validation should happen as early as possible, ideally in `setContactPoints` to not allow invalid values to get into the factory bean in the first place. `afterPropertiesSet()` is kind-of after the fact already and only good to check if required values weren't configured.

As discussed for backward compatibility, we should leniently accept host or host and port. We also need to check if the host contains an IPv6 as `:` is used in IPv6 addresses, so we'd need to check whether the address is using the resource identifier format with square brackets (`[2001:db8:85a3:8d3:1319:8a2e:370:7348]:443`) so that we correctly identify the port.

tomekl007:
> Backward compatibility is given as `contactPoints` expects one or more hosts without a port. If we discover a hostname without a port, then we use the `port` value that was set on the factory bean. If a contact point was entered with a port, then the port of the contact point has precedence over `setPort(…)`. It makes sense to reflect this behavior in the Javadocs of both setters.

What about the situation, in which the user will provide a port for some contact-points, and forget adding it for other settings:
`a:1000,b:2000,c`
are you suggesting to use the `DEFAULT_PORT` for the hostname `c`?
If yes, then we cannot do much validation here because both `host:port` and `host` are correct. 
Are you sure that we should allow such partially configured `contact-points` and append `port` implicitly to some hosts? I imagine that it may lead to misconfiguration that will be hard to catch by the end-users.

mp911de:
For `a:1000,b:2000,c`, `c` would apply the port from `setPort` (during `afterPropertiesSet`), basically doing the same as today. Therefore, the change would extend today's functionality so that `contactPoints` becomes more capable.

Improving on configuration semantics makes sense. We learned from the past that having two properties (contactPoints vs. contactPointsWithPorts) is more confusing and a bigger source of misconfiguration than injecting contextual information as one has always to look up the documentation to be sure about semantics.

With specifying a single property for hosts/contact points, one has only a single source of truth. If a host uses a port-less form, then we enhance details from a fallback that is strictly tied to a single attribute (port).

What we could do to improve here is introducing another overload accepting `InetSocketAddress` (`setContactPoints(Collection<InetSocketAddress>)`which doesn't leave any ambiguity.

tomekl007:
> What we could do to improve here is introducing another overload accepting `InetSocketAddress` (`setContactPoints(Collection<InetSocketAddress>)`which doesn't leave any ambiguity.

I agree that adding `setContactPoints(Collection<InetSocketAddress>)` seems like a sane solution that is more flexible.
Do you think that in such a solution, the current `setContactPoints` and `setPort` should be left unchanged and the new  (`setContactPoints(Collection<InetSocketAddress>)`) would be an only way for users to provide port per host?
What if both versions will be set by the user? What if only a new overloaded version?
It can be solved by making the overloaded version having priority over `setContactPoints(Stirng)` and `setPort(Int)`, wdyt?


The one downside of this solution is that we are pushing the responsibility of creating resolved vs unresolved `InetSocket` address to end-users, are you ok with that?

mp911de:
I'd like to keep a consistent behavior between Spring Data and Spring Boot. Since Boot is already applying port defaulting (see https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraProperties.java#L53-L62) if the contact point does not specify a port.

`setContactPoints(Collection<InetSocketAddress>)` overrides what was set with `setContactPoints(String)` and vice versa. From a configuration model perspective, the factory bean defines a property `contactPoints` that can be either set as `String` or as `Collection<InetSocketAddress>`. The last invocation to `setContactPoints(…)` wins (as it is with regular Java beans) and gets applied in `afterPropertiesSet()`.

tomekl007:
> I'd like to keep a consistent behavior between Spring Data and Spring Boot. Since Boot is already applying port defaulting (see https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraProperties.java#L53-L62) if the contact point does not specify a port.
> 
> `setContactPoints(Collection<InetSocketAddress>)` overrides what was set with `setContactPoints(String)` and vice versa. From a configuration model perspective, the factory bean defines a property `contactPoints` that can be either set as `String` or as `Collection<InetSocketAddress>`. The last invocation to `setContactPoints(…)` wins (as it is with regular Java beans) and gets applied in `afterPropertiesSet()`.

I agree with the argument about consistency between APIs. I've changed the behavior to be consistent with spring boot in: 
https://github.com/spring-projects/spring-data-cassandra/pull/178/commits/094b916e3be57e44d28af397e0eee6bd1b2f7584
To be fully consistent, I also ported and adapted logic from Spring Boot `CassandraAutoConfigiuration#mapContactPoints` and `#formatContactPoint`.

mp911de:
> What do you mean by it can be set to two different types? The type of contactPoints can be either String OR Collection<InetSocketAddress) not both.

We define two methods: `setContactPoints(String)` and `setContactPoints(Collection<InetSocketAddress>)` to configure the `contactPoints` property. From a property perspective, calling a setter replaces entirely the previous value so there's no ambiguity to begin with.

We make use of the fact that `contactPoints` can be represented either as `String` or as collection of `InetSocketAddress`.

Inside of the factory bean we can represent the actual value as `IntFunction<Collection<InetSocketAddress>>` that is being called with the port as argument.

Calling `setContactPoints(Collection<InetSocketAddress> contactPoints)` sets the value to `unusedPort -> contactPoints` while calling `setContactPoints(String contactPoints)` sets the value to `port -> createInetSocketAddresses(contactPoints, port)`

tomekl007:
> > What do you mean by it can be set to two different types? The type of contactPoints can be either String OR Collection<InetSocketAddress) not both.
> 
> We define two methods: `setContactPoints(String)` and `setContactPoints(Collection<InetSocketAddress>)` to configure the `contactPoints` property. From a property perspective, calling a setter replaces entirely the previous value so there's no ambiguity to begin with.
> 
> We make use of the fact that `contactPoints` can be represented either as `String` or as collection of `InetSocketAddress`.
> 
> Inside of the factory bean we can represent the actual value as `IntFunction<Collection<InetSocketAddress>>` that is being called with the port as argument.
> 
> Calling `setContactPoints(Collection<InetSocketAddress> contactPoints)` sets the value to `unusedPort -> contactPoints` while calling `setContactPoints(String contactPoints)` sets the value to `port -> createInetSocketAddresses(contactPoints, port)`

Good idea, I had the impression that we don't want to modify the internal `contactPoints` variable but since it is possible I've changed the implementation according to your suggestion.

mp911de:

PR #179, opened 2020-08-13
DATACASS-751: Allow keyspace customization for CassandraPersistentEntity

tomekl007:

todo:
- [ ] adapt Events (i.e `BeforeDeleteEvent`, `BeforeSaveEvent`, `AfterSaveEvent`, etc) to `TableCoordinates`
- [ ] cover `TableCoordinates` with keyspace with more IT and UnitTests
- [ ] consider `CassandraTemplate#getMapper` to work per `TableCoordinates` (not only `tableName`)
- [ ] consider moving TableCoordinates to `CassandraPersistentEntity#getTableCoordinates()` AND `BasicCassandraPersistentEntity`

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/AsyncCassandraTemplate.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/AsyncCassandraTemplate.java
@@ -69,6 +69,7 @@ import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.DriverException;
 import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
 import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
+import com.datastax.oss.driver.api.core.cql.ResultSet;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.datastax.oss.driver.api.core.cql.SimpleStatement;
 import com.datastax.oss.driver.api.core.cql.Statement;
@@ -94,6 +95,7 @@ import com.datastax.oss.driver.api.querybuilder.update.Update;
  *
  * @author Mark Paluch
  * @author John Blum
+ * @author Tomasz Lelek
  * @see org.springframework.data.cassandra.core.AsyncCassandraOperations
  * @since 2.0
  */
@@ -122,7 +124,7 @@ public class AsyncCassandraTemplate
 	 *
 	 * @param session {@link CqlSession} used to interact with Cassandra; must not be {@literal null}.
 	 * @see CassandraConverter
-	 * @see Session
+	 * @see CqlSession
 	 */
 	public AsyncCassandraTemplate(CqlSession session) {
 		this(session, newConverter());
@@ -136,7 +138,7 @@ public class AsyncCassandraTemplate
 	 * @param converter {@link CassandraConverter} used to convert between Java and Cassandra types; must not be
 	 *          {@literal null}.
 	 * @see CassandraConverter
-	 * @see Session
+	 * @see CqlSession
 	 */
 	public AsyncCassandraTemplate(CqlSession session, CassandraConverter converter) {
 		this(new DefaultSessionFactory(session), converter);
@@ -150,7 +152,7 @@ public class AsyncCassandraTemplate
 	 * @param converter {@link CassandraConverter} used to convert between Java and Cassandra types; must not be
 	 *          {@literal null}.
 	 * @see CassandraConverter
-	 * @see Session
+	 * @see CqlSession
 	 */
 	public AsyncCassandraTemplate(SessionFactory sessionFactory, CassandraConverter converter) {
 		this(new AsyncCqlTemplate(sessionFactory), converter);
@@ -164,7 +166,7 @@ public class AsyncCassandraTemplate
 	 * @param converter {@link CassandraConverter} used to convert between Java and Cassandra types; must not be
 	 *          {@literal null}.
 	 * @see CassandraConverter
-	 * @see Session
+	 * @see CqlSession
 	 */
 	public AsyncCassandraTemplate(AsyncCqlTemplate asyncCqlTemplate, CassandraConverter converter) {
 
@@ -267,6 +269,10 @@ public class AsyncCassandraTemplate
 
 	private CqlIdentifier getTableName(Class<?> entityClass) {
 		return getEntityOperations().getTableName(entityClass);
+	}
+
+	private CqlIdentifier getKeyspace(Class<?> entityClass) {
+		return getEntityOperations().getKeyspace(entityClass);
 	}
 
 	// -------------------------------------------------------------------------
@@ -482,7 +488,7 @@ public class AsyncCassandraTemplate
 
 		Assert.notNull(entityClass, "Entity type must not be null");
 
-		return doCount(Query.empty(), entityClass, getTableName(entityClass));
+		return doCount(Query.empty(), entityClass, getKeyspace(entityClass), getTableName(entityClass));
 	}
 
 	/* (non-Javadoc)
@@ -494,13 +500,14 @@ public class AsyncCassandraTemplate
 		Assert.notNull(query, "Query must not be null");
 		Assert.notNull(entityClass, "Entity type must not be null");
 
-		return doCount(query, entityClass, getTableName(entityClass));
+		return doCount(query, entityClass, getKeyspace(entityClass), getTableName(entityClass));
 	}
 
-	ListenableFuture<Long> doCount(Query query, Class<?> entityClass, CqlIdentifier tableName) {
+	ListenableFuture<Long> doCount(Query query, Class<?> entityClass, @Nullable CqlIdentifier keyspace,
+			CqlIdentifier tableName) {
 
 		StatementBuilder<com.datastax.oss.driver.api.querybuilder.select.Select> countStatement = getStatementFactory()
-				.count(query, getRequiredPersistentEntity(entityClass), tableName);
+				.count(query, getRequiredPersistentEntity(entityClass), keyspace, tableName);
 
 		SimpleStatement statement = countStatement.build();
 
@@ -521,7 +528,7 @@ public class AsyncCassandraTemplate
 		CassandraPersistentEntity<?> entity = getRequiredPersistentEntity(entityClass);
 
 		StatementBuilder<com.datastax.oss.driver.api.querybuilder.select.Select> select = getStatementFactory()
-				.selectOneById(id, entity, entity.getTableName());
+				.selectOneById(id, entity, entity.getKeyspace(), entity.getTableName());
 
 		return new MappingListenableFutureAdapter<>(getAsyncCqlOperations().queryForResultSet(select.build()),
 				resultSet -> resultSet.one() != null);
@@ -537,7 +544,8 @@ public class AsyncCassandraTemplate
 		Assert.notNull(entityClass, "Entity type must not be null");
 
 		StatementBuilder<com.datastax.oss.driver.api.querybuilder.select.Select> select = getStatementFactory()
-				.select(query.limit(1), getRequiredPersistentEntity(entityClass), getTableName(entityClass));
+				.select(query.limit(1), getRequiredPersistentEntity(entityClass), getKeyspace(entityClass),
+						getTableName(entityClass));
 
 		return new MappingListenableFutureAdapter<>(getAsyncCqlOperations().queryForResultSet(select.build()),
 				resultSet -> resultSet.one() != null);
@@ -554,7 +562,7 @@ public class AsyncCassandraTemplate
 
 		CassandraPersistentEntity<?> entity = getRequiredPersistentEntity(entityClass);
 		CqlIdentifier tableName = entity.getTableName();
-		StatementBuilder<Select> select = getStatementFactory().selectOneById(id, entity, tableName);
+		StatementBuilder<Select> select = getStatementFactory().selectOneById(id, entity, entity.getKeyspace(), tableName);
 		Function<Row, T> mapper = getMapper(entityClass, entityClass, tableName);
 
 		return new MappingListenableFutureAdapter<>(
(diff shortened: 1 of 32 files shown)


tomekl007:
I am opening this as a draft PR to discuss if the changes are in the right direction.

Right now I modified only `QueryBuilder.selectFrom` to make it work for keyspace and table. The changes are backward compatible, and the keyspace can be null. 
The keyspace is propagated from the `CassandraMappingContext`:
https://github.com/spring-projects/spring-data-cassandra/pull/179/files#diff-2e567c3232b51d1962846f76702e5423R281
and is tied to the`CassandraPersistentEntity.java` as discussed in the JIRA ticket (https://github.com/spring-projects/spring-data-cassandra/pull/179/files#diff-f553f0af64a53e7dd83c965ec8362159R206)
@mp911de let me know wdyt if this is going in the right direction.
If yes, I will proceed in the same way with:
- QueryBuilder.truncate
- QueryBuilder.insertInto
- QueryBuilder.update
- QueryBuilder.deleteFrom
methods

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/AsyncCassandraTemplate.java
```diff
@@ -482,7 +488,7 @@ private CqlIdentifier getTableName(Class<?> entityClass) {
 
 		Assert.notNull(entityClass, "Entity type must not be null");
 
-		return doCount(Query.empty(), entityClass, getTableName(entityClass));
+		return doCount(Query.empty(), entityClass, getKeyspace(entityClass), getTableName(entityClass));
```
mp911de, line 491: Since the table name and keyspace name are used together, it would make sense to introduce a value object that encapsulates both pieces of information. How about introducing an object `TableName` that contains the table name and an optional keyspace name?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/ExecutableSelectOperation.java
```diff
@@ -102,6 +102,26 @@
 		 */
 		SelectWithProjection<T> inTable(CqlIdentifier table);
 
+		/**
```
mp911de, line 105: This interface is implemented by Spring Data Cassandra only, so we don't need to add default methods for backward compatibility.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/CassandraMappingContext.java
```diff
@@ -73,6 +74,8 @@
 
 	private NamingStrategy namingStrategy = NamingStrategy.INSTANCE;
 
+	private @Nullable CqlIdentifier keyspace;
```
mp911de, line 77: We don't need a keyspace here.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/BasicCassandraPersistentEntity.java
```diff
@@ -226,6 +226,12 @@ public CqlIdentifier getTableName() {
 		return Optional.ofNullable(this.tableName).orElseGet(this::determineTableName);
 	}
 
+	@Nullable
+	@Override
+	public CqlIdentifier getKeyspace() {
```
mp911de, line 231: Let's turn that method into an `Optional`. We should make use on the metadata-level of improved null-safety.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/CassandraPersistentEntity.java
```diff
@@ -50,6 +52,12 @@
 	 */
 	CqlIdentifier getTableName();
 
+
+	/**
+	 * Returns the keyspace to which the entity shall be persisted.
+	 */
+	@Nullable CqlIdentifier getKeyspace();
```
mp911de, line 59: Let's turn this method's return type into `Optional` and rename it to `getKeyspaceName`

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/CassandraPersistentEntity.java
```diff
@@ -70,6 +78,7 @@ default void setTableName(org.springframework.data.cassandra.core.cql.CqlIdentif
 	 */
 	void setTableName(CqlIdentifier tableName);
 
+
```
mp911de, line 81: As we have a `setTableName(…)` option, we could introduce `setKeyspaceName(…)` for consistency.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/NamingStrategy.java
```diff
@@ -53,6 +55,11 @@ default String getTableName(CassandraPersistentEntity<?> entity) {
 		return entity.getType().getSimpleName();
 	}
 
+	default Optional<CqlIdentifier> getKeyspace(CassandraPersistentEntity<?> entity) {
```
mp911de, line 58: Let's rename this to `getKeyspaceName()`

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/AsyncCassandraTemplate.java
tomekl007: I see two propositions from you: 
- create a value object
- leave it as it is, but rename `getKeysapce` to `getKeyspaceName`
So which one should I pick?

mp911de: Let's go with `getKeyspaceName` to align with `getTableName`.

tomekl007: So regarding value objects, do you mean that it should be encapsulated and used internally? I.e. passed to the `doCount` whereas the public API should have `getKeyspaceName` and `getTableName`, right?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/BasicCassandraPersistentEntity.java
tomekl007: The question is at which level we should transform this `Optional` back to a value.
I.e: when following your suggestion I need to change all callers and this will be propagated up to specific method, i.e:
```
public StatementBuilder<Select> count(Query query, CassandraPersistentEntity<?> entity,
           Optional<CqlIdentifier> keyspace, CqlIdentifier tableName) //
``` 
but it will generate the `Optional<CqlIdentifier> used as a type for parameter keyspace` warning; this is the reason why I did use `@Nullable` instead of an `Optional`. 
Are we ok with a warning?
If not, but If we still want to change those methods to return `Optional` (`CassandraPersistemtEntity#getKeyspace` AND `BasicCassandraPersistemtEntity#getKeyspace`) then we need to transform from `Optional` to a value in every caller, i.e:
```
public StatementBuilder<Select> count(Query query, CassandraPersistentEntity<?> persistentEntity) {
               // ....
               return count(query, persistentEntity, persistentEntity.getKeyspace().orElse(null), persistentEntity.getTableName());
	}
```
and this `persistentEntity.getKeyspace().orElse(null)` will be propagated in a lot of places in code so that is far from ideal.

mp911de: > The question is at which level we should transform this Optional back to a value.

Ideally, all methods accepting `CqlIdentifier` as table name get overloaded (if they are public) to accept a `TableName`/`TableCoordinates`/`CqlTableName` object so that keyspace retrieval happens as late as possible and in as little places as possible.

Something along the lines of:


```java
private static StatementBuilder<Select> createSelectAndOrder(List<Selector> selectors, CqlTableName from,
		Filter filter, Sort sort) {

	Select select;

	if (selectors.isEmpty()) {
		select = from.map((keyspace, tableName) -> QueryBuilder.selectFrom(keyspace, tableName)).orElseGet(tableName -> QueryBuilder.selectFrom(tableName)).all();
```

which was previously:

```java
private static StatementBuilder<Select> createSelectAndOrder(List<Selector> selectors, CqlIdentifier from,
		Filter filter, Sort sort) {

	Select select;

	if (selectors.isEmpty()) {
		select = QueryBuilder.selectFrom(from).all();
```

As mentioned, I'm not 100% convinced about naming the value object `TableName`, so happy to discuss suggestions.

On the `PersistentEntity`-level it makes sense to keep getters for `keyspaceName` and `tableName`.

tomekl007: Regarding the naming, I think that it would be beneficial to have the `Keysapce` in the name of this new type.
So maybe just `KeyspaceTable` OR  `KeyspaceTableName`?

mp911de: Rather than enumerating the properties of which the information consists, we should name the thing that it describes.

tomekl007: In that case, I think that `TableCoordinates` describes it best.

mp911de: Sounds good. Then we can start working on that one as value object that consists of the actual name and an optional keyspace name. Typically, we hide constructors and provide factory methods along the lines of `of(CqlIdentifier tableName, Optional<CqlIdentifier> keyspaceName)`. We also should consider how the contents gets used. Besiders getters it would make sense to allow a somewhat functional approach that allows mapping of both information (if present) and fall back to the table name if the keyspace isn't specified.

I'll be off for about two weeks, so feel free to experiment with what feels best.

tomekl007: regarding mapping, I think that this is not necessary. All `queryBuilder` methods are accepting null keyspace (and ignore it if it's null):
```
@NonNull
  public static InsertInto insertInto(
      @Nullable CqlIdentifier keyspace, @NonNull CqlIdentifier table) {
```
so we can just pass `null`:
```
QueryBuilder.insertInto(tableCoordinates.getKeyspaceName().orElse(null), tableCoordinates.getTableName())
```
without a need for `optional.map`.

tomekl007:
One more question @mp911de.
The logic for do*Versioned is emitting events like for example:
`maybeEmitEvent(new AfterSaveEvent<>(entityToSave, tableCoordinates.getTableName()));`
and:
```
public AfterSaveEvent(E source, CqlIdentifier tableName) {
     super(source, tableName);
}
```
It seems that we would need to modify events to send TableCoordinates instead of only `tableName`, right?

mp911de:
Let's keep events and entity callbacks for a future step in mind so we don't change these yet.

tomekl007 (commented):
re: `getMapper()` method and it's invocations.
The mapper currently is created per `Table` as we discussed. It creates one problem.
For example, looking at a place where `TableCoordinates` are used:
```
<T> List<T> doSelect(Query query, Class<?> entityClass, TableCoordinates tableCoordinates,
			Class<T> returnType) {
```
and the actual logic of getting mapper is happening per table:
```
	Function<Row, T> mapper = getMapper(entityClass, returnType, tableCoordinates.getTableName());
```
So it basically means that `TableCoordinates.keyspaceName` is ignored for all `CassandraTemplate` methods that are using the mapper.
It means that the behavior of methods that are using mapper vs those that are using statements directly (created by the `StatementFactory`) is different. 

It may lead to in-deterministic behavior if the client has a table with the same name in two different keyspaces. The client will be able to execute per ks queries using Statements, but when it comes to a mapper, only one table could be retrieved and the `keyspace` is inherited from the `CqlSession`. Are we ok with this behavior? Or should we create mapper per keyspace-table as well?

tomekl007:
> Let's keep events and entity callbacks for a future step in mind so we don't change these yet.

What do you mean by the future step? Do you want to do it in a separate ticket (PR)?

tomekl007:
question regarding testing:
I would like to add tests for all `*OperationSupport` classes that are working now per TableCoordinates, and also have a new method: 
`UpdateWithQuery inTable(CqlIdentifier keyspaceName, CqlIdentifier tableName)`
So for example adding new tests to `ExecutableUpdateOperationSupportIntegrationTests`.
To test the new per `kespace.table` behavior, I would like to use `CassandraAdminTemplate` to create a table in the specific keyspace. Right now, the `CassandraAdminTemplate` is exposing methods that are working only for table and keyspace cannot be specified explicitly, i.e: 'public void dropTable(CqlIdentifier tableName)`, `public void createTable(boolean ifNotExists, CqlIdentifier tableName...`. I think that we should consider adding new methods to `CassandraAdminTemplate` that are working per `keyspace/table`. They will be used to write good integration tests, but also end-users may find it useful. wdyt?

mp911de:
>> Let's keep events and entity callbacks for a future step in mind so we don't change these yet.
>
> What do you mean by the future step? Do you want to do it in a separate ticket (PR)?

Yes, I'd like to not introduce another level of complexity with this pull request but rather to postpone changes to events and entity callbacks to after this PR is merged.

mp911de:
It makes sense to extend `CassandraAdminTemplate` by introducing overloads accepting `TableCooordinates`. Going further, I'm wondering about UDT's since they have a name and they exist in a keyspace. Is it possible to do something similar as with tables, so that a UDT gets created and referenced in a different keyspace?

I'm also wondering whether our `TableCoordinates`-approach is the right thing to do. In Spring Data JDBC, we faced a similar request, to prefix tables with a schema name and we've built [`SqlIdentifier`](https://github.com/spring-projects/spring-data-jdbc/blob/master/spring-data-relational/src/main/java/org/springframework/data/relational/core/sql/SqlIdentifier.java) that can be either a simple (table/column) or a composite identifier (schema with table).

In the Cassandra case, we're referencing to the Driver `CqlIdentifier`. I'm a bit hesitant to subclass `CqlIdentifer` for representing table name and keyspace name as the type isn't under our control and the Driver could decide to make this type final. Additionally, it's a class and not an interface which makes it a bit more complex to create subclasses.

The gist here is, if `CqlIdentifier` would be a bit more friendly to extend, then we could simplify this PR a lot more. Paging @adutra for further discussion.

mp911de:
Superseded as per https://github.com/spring-projects/spring-data-cassandra/issues/921#issuecomment-2250284092. Thank you for your contribution that inspired the design of customizing keyspaces.

PR #181, opened 2020-11-13
DATACASS-825 - Implements deleteAllById(Iterable<ID> ids).

schauder:
Merge after DATACMNS-800

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/SimpleCassandraRepository.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/SimpleCassandraRepository.java
@@ -44,6 +44,7 @@ import com.datastax.oss.driver.api.querybuilder.insert.Insert;
  * @author Matthew T. Adams
  * @author Mark Paluch
  * @author John Blum
+ * @author Jens Schauder
  * @see org.springframework.data.cassandra.repository.CassandraRepository
  */
 public class SimpleCassandraRepository<T, ID> implements CassandraRepository<T, ID> {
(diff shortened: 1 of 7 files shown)


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/SimpleCassandraRepository.java
```diff
@@ -247,6 +248,14 @@ public void deleteAll(Iterable<? extends T> entities) {
 		entities.forEach(this.operations::delete);
 	}
 
+	@Override
+	public void deleteAllById(Iterable<? extends ID> ids) {
+
+		Assert.notNull(ids, "The given Iterable of ids must not be null");
+
+		ids.forEach(id -> operations.deleteById(id, entityInformation.getJavaType()));
```
mp911de, line 256: We should use the same semantics as with `findAllById(…)` regarding the query mechanism to avoid individual database queries for each ID.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/SimpleReactiveCassandraRepository.java
```diff
@@ -304,6 +305,14 @@ public SimpleReactiveCassandraRepository(CassandraEntityInformation<T, ID> metad
 		return Flux.fromIterable(entities).flatMap(this.operations::delete).then();
 	}
 
+	@Override
+	public Mono<Void> deleteAllById(Iterable<? extends ID> ids) {
+
+		Assert.notNull(ids, "The given Iterable of entities must not be null");
+
+		return Flux.fromIterable(ids).flatMap(this::deleteById).then();
```
mp911de, line 313: We should use the same semantics as with `findAllById(…)` regarding the query mechanism to avoid individual database queries for each ID.

mp911de (changes requested):

mp911de (approved):

mp911de:

PR #1108, opened 2021-03-28
DATACASS-812 - Added schema generation support for static columns.

### LINKED ISSUE #978: Creating columns with a single value (static column) [DATACASS-812]
**[lmcdasi](https://jira.spring.io/secure/ViewProfile.jspa?name=lmcdasi)** opened **[DATACASS-812](https://jira.spring.io/browse/DATACASS-812?redirect=false)** and commented

In the DataStax driver manual, there is a possibility to create columns as static.

https://docs.datastax.com/en/dse/6.0/cql/cql/cql_using/refStaticCol.html

In the driver code, the DefaultAlterTable seems to support the addition of static columns through: columnsToAddStatic

I'm wondering if we can achieve the same with the framework & if not is it possible introduce this functionality ?

Thx

---
No further details from [DATACASS-812](https://jira.spring.io/browse/DATACASS-812?redirect=false)

alex-ninja:
Closes #978

### Overview of the changes
This PR addresses #978 / DATACASS-812 feature request. The main goal is to support single-value (static) columns. Instead of the suggested in the original issue approach based on an additional attribute in `@Column` annotation (e.g. `@Column(name = "x", isStatic = true)`, I have added a separate `@StaticColumn` annotation. It seems to be more flexible because the `@Column` annotation requires to set a column name (which might be inconvenient when an end user does not want to override the default naming).

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/ColumnSpecification.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/ColumnSpecification.java
@@ -29,10 +29,10 @@ import com.datastax.oss.driver.api.core.type.DataType;
 /**
  * Object to configure a CQL column specification.
  * <p/>
- * Use {@link #name(String)} and {@link #type(String)} to set the name and type of the column, respectively. To specify
+ * Use {@link #name(String)} and {@link #type(DataType)} to set the name and type of the column, respectively. To specify
  * a clustered {@code PRIMARY KEY} column, use {@link #clustered()} or {@link #clustered(Ordering)}. To specify that the
  * {@code PRIMARY KEY} column is or is part of the partition key, use {@link #partitioned()} instead of
- * {@link #clustered()} or {@link #clustered(Ordering)}.
+ * {@link #clustered()} or {@link #clustered(Ordering)}. To specify {@code STATIC} column, use {@link #staticColumn()}.
  *
  * @author Matthew T. Adams
  * @author Alex Shvid
@@ -52,6 +52,8 @@ public class ColumnSpecification {
 	private @Nullable PrimaryKeyType keyType;
 
 	private @Nullable Ordering ordering;
+
+	private boolean isStatic;
 
 	private ColumnSpecification(CqlIdentifier name) {
 		this.name = name;
@@ -178,6 +180,19 @@ public class ColumnSpecification {
 		return this;
 	}
 
+	/**
+	 * Identifies this column as a static column. Sets the column's {@link #isStatic} to {@literal true}.
+	 *
+	 * @return this
+	 * @since 3.2
+	 */
+	public ColumnSpecification staticColumn() {
+
+		this.isStatic = true;
+
+		return this;
+	}
+
 	public CqlIdentifier getName() {
 		return name;
 	}
@@ -197,6 +212,10 @@ public class ColumnSpecification {
 		return ordering;
 	}
 
+	public boolean isStatic() {
+		return isStatic;
+	}
+
 	public String toCql() {
 		return toCql(new StringBuilder()).toString();
 	}
@@ -211,7 +230,10 @@ public class ColumnSpecification {
 	 */
 	@Override
 	public String toString() {
-		return toCql(new StringBuilder()).append(" /* keyType=").append(keyType).append(", ordering=").append(ordering)
+		return toCql(new StringBuilder()).append(" /* ")
+				.append("keyType=").append(keyType).append(", ")
+				.append("ordering=").append(ordering).append(", ")
+				.append("isStatic=").append(isStatic)
 				.append(" */ ").toString();
 	}
 }
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/TableSpecification.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/TableSpecification.java
@@ -59,6 +59,11 @@ public class TableSpecification<T> extends TableOptionsSpecification<TableSpecif
 	 */
 	private List<ColumnSpecification> nonKeyColumns = new ArrayList<>();
 
+	/**
+	 * List of static columns.
+	 */
+	private List<ColumnSpecification> staticColumns = new ArrayList<>();
+
 	protected TableSpecification(CqlIdentifier name) {
 		super(name);
 	}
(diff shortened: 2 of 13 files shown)


alex-ninja:
Hi @mp911de !
Could you please briefly take a look and provide an initial feedback on this PR.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/StaticColumn.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Annotation to mark a field static.
+ *
+ * @author Aleksei Zotov
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(value = { ElementType.ANNOTATION_TYPE, ElementType.FIELD, ElementType.METHOD })
+public @interface StaticColumn {
```
mp911de, line 32: Instead of introducing another annotation type, it seems that we should rather extend `@Column` and probably `@PrimaryKeyColumn` by adding a `staticColumn` or `isStatic` attribute.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/ColumnSpecification.java
```diff
@@ -178,6 +180,18 @@ public ColumnSpecification ordering(Ordering ordering) {
 		return this;
 	}
 
+	/**
+	 * Identifies this column as a static column. Sets the column's {@link #isStatic} to {@literal true}.
+	 *
+	 * @return this
```
mp911de, line 186: Nit: `@since 3.2`

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/TableSpecification.java
```diff
@@ -82,7 +87,29 @@ public T column(String name, DataType type) {
...
+	}
+
+	/**
+	 * Adds the given static column to the table. Must be specified after all primary key columns.
+	 *
+	 * @param name The column name; must be a valid unquoted or quoted identifier without the surrounding double quotes,
+	 *          must not be {@literal null}.
+	 * @param type The data type of the column, must not be {@literal null}.
+	 */
+	public T staticColumn(String name, DataType type) {
```
mp911de, line 100: Nit: `@since 3.2`

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/TableSpecification.java
```diff
@@ -82,7 +87,29 @@ public T column(String name, DataType type) {
...
+	}
+
+	/**
+	 * Adds the given static column to the table. Must be specified after all primary key columns.
+	 *
+	 * @param name The column name; must be a valid unquoted or quoted identifier without the surrounding double quotes,
+	 *          must not be {@literal null}.
+	 * @param type The data type of the column, must not be {@literal null}.
+	 */
+	public T staticColumn(CqlIdentifier name, DataType type) {
```
mp911de, line 111: Nit: `@since 3.2`

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/TableSpecification.java
```diff
@@ -317,4 +350,12 @@ protected T column(CqlIdentifier name, DataType type, Optional<PrimaryKeyType> o
 	public List<ColumnSpecification> getNonKeyColumns() {
 		return Collections.unmodifiableList(this.nonKeyColumns);
 	}
+
+	/**
+	 * Returns an unmodifiable list of static columns.
+	 */
+	@Override
+	public List<ColumnSpecification> getStaticColumns() {
```
mp911de, line 358: Nit: `@since 3.2`

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/CassandraPersistentProperty.java
```diff
@@ -179,4 +179,13 @@ default boolean isEmbedded() {
 	@Nullable
 	AnnotatedType findAnnotatedType(Class<? extends Annotation> annotationType);
 
+
+	/**
+	 * Whether the property is a static column.
+	 * @since 3.2
+	 */
+	default boolean isStaticColumn() {
```
mp911de, line 187: Annotation lookup can be somewhat expensive. It makes sense to implement the lookup directly in `BasicCassandraPersistentProperty` and update `CachingCassandraPersistentProperty` to cache the outcome whether that's a static column.

mp911de (changes requested):
Thanks for your pull request. I left a few comments. Design-wise, we should rather update existing annotations instead of introducing new ones. It aligns better with the purpose of annotations defining their purpose and by doing so we can exclude static columns from being used with `@PrimaryKey`. Since I'm not sure whether static columns make sense in the context of clustering keys, I'm not sure whether we should extend `@PrimaryKeyColumn` as well.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/ColumnSpecification.java
alex-ninja: Done

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/TableSpecification.java
alex-ninja: Done

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/TableSpecification.java
alex-ninja: Done

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/cql/keyspace/TableSpecification.java
alex-ninja: Done

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/CassandraPersistentProperty.java
alex-ninja: Good catch! I've made the corresponding fix and updated `BasicCassandraPersistentTupleProperty` with `false` since it cannot be static. Additionally, I've added a separate test for a static tuple field.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/StaticColumn.java
alex-ninja: As proposed, I've added `isStatic` attribute to `@Column` annotation. There is no need to add that to `@PrimaryKeyColumn`  annotation because primary columns (partitioned/clustered) cannot be static ([reference to documentation](https://docs.datastax.com/en/dse/6.8/cql/cql/cql_using/refStaticCol.html)). There is a separate check for that:
```
Assert.isTrue(!(optionalKeyType.isPresent() && isStatic),"PrimaryKey must not be static");
```

alex-ninja:
@mp911de Thanks for valuable feedback! I've addressed all the comments and squashed everything into a single commit (as stated in the contribution guide).

alex-ninja:
@mp911de 
Were you able to check the PR? I'm sorry for being slightly impatient, just want to confirm it is not missing from your radar.

mp911de:
No need to worry. This PR is still being considered for the upcoming GA release.

mp911de:

Issue #1196, opened 2021-11-19
Iteration order can change when using `Columns.from(…)`


LeoYimingLi:
- The flakiness in ```ColumnsUnitTests#shouldCreateFromColumns```https://github.com/spring-projects/spring-data-cassandra/blob/880787ff8841b461c3bff6d0755a485171867a59/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/core/query/ColumnsUnitTests.java#L47-L53 is due to the function ```Columns from(String... columnNames)``` where ```HashMap``` is used and the order of map is not guaranteed. 
https://github.com/spring-projects/spring-data-cassandra/blob/880787ff8841b461c3bff6d0755a485171867a59/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/query/Columns.java#L69-L78

- We can use ```LinkedHashMap``` to replace ```HashMap``` to keep the order deterministic.  Tentative PR to fix the flaky test: https://github.com/LeoYimingLi/spring-data-cassandra/pull/1/files

- version of spring data
  SHA: 880787ff8841b461c3bff6d0755a485171867a59

- JVM version
  > openjdk version "1.8.0_292"
OpenJDK Runtime Environment (build 1.8.0_292-8u292-b10-0ubuntu1~18.04-b10)
OpenJDK 64-Bit Server VM (build 25.292-b10, mixed mode)

- The way to reproduce the flaky test failure:
  - 0.Environment setup: Maven 3.6.0 and Java 1.8.0_292
  - 1.clone the repo:
  ```
  git clone https://github.com/spring-projects/spring-data-cassandra
  cd spring-data-cassandra
  git checkout 880787ff8841b461c3bff6d0755a485171867a59
  ```
  - 2.run with [Nondex tool](https://github.com/TestingResearchIllinois/NonDex) (which explores different behaviors of under-determined APIs and reports test failures)
  ```
  mvn install -pl spring-data-cassandra -am -DskipTests
  mvn -pl spring-data-cassandra edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=org.springframework.data.cassandra.core.query.ColumnsUnitTests#shouldCreateFromColumns
  ```
  - 3.And we can get the test failure screenshot:
  ![image](https://user-images.githubusercontent.com/46290389/142703963-574693cc-f1df-4df7-81e2-4856016ec90a.png)

PR #1198, opened 2021-11-22
fix flaky test in ColumnsUnitTests

CharlesZKQ:
In `org.springframework.data.cassandra.core.query.ColumnsUnitTests`, the test `shouldCreateFromColumns()` is flaky due to the non-deterministic property of `HashMap` (please refer [document](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html)). it is internally invoked by `Columns.from()`. I fixed it checking whether it contains the name of these two columns instead of comparing the whole string, which doesn't affected by the order of string generated by non-determined `Columns.from()`.


mp911de:
Superseded by #1196.

PR #1199, opened 2021-11-22
fix flaky test in CriteriaUnitTests

CharlesZKQ:
In `org.springframework.data.cassandra.core.query.CriteriaUnitTests`, the tests `shouldCreateIsInSet()` is  flaky due to the non-deterministic property of `HashSet` (please refer [document](https://docs.oracle.com/javase/7/docs/api/java/util/HashSet.html)).I fixed by changing the `HashSet` to `LinkedHashSet`, which guarantees the order of value added into the `set` and eliminates the flakiness of this test.


mp911de:
Thank you for your contribution. That's merged now.

PR #1200, opened 2021-11-22
fix flaky test in PartTreeCassandraQueryUnitTests

CharlesZKQ:
In `org.springframework.data.cassandra.repository.query.PartTreeCassandraQueryUnitTests`, the test `usesProjectionQueryHiddenField()` is flaky due to the non-deterministic property of `HashMap` (please refer [document](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html)). it is internally invoked by `createQuery()` in method `deriveQueryFromMethod()`. I fixed by comparing the generated actual string with all possible order of query strings, if it satisfies either one of the possible solution, it will pass the test.


mp911de:
Superseded by #1196.

PR #1201, opened 2021-11-22
Fix flaky test in ReactivePartTreeCassandraQueryUnitTests

CharlesZKQ:
In `org.springframework.data.cassandra.repository.query.ReactivePartTreeCassandraQueryUnitTests`, the test `usesDynamicProjection()` is flaky due to the non-deterministic property of `HashMap` (please refer [document](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html)). it is internally invoked by `createQuery()` in method `deriveQueryFromMethod()`. I fixed by comparing the generated actual string with all possible order of query strings, if it satisfies either one of the possible solution, it will pass the test. This PR is similar to [pull/1200](https://github.com/spring-projects/spring-data-cassandra/pull/1200).


mp911de:
Superseded by #1196.

Issue #1253, opened 2022-04-14
Don't refresh schema metadata if it's disabled by the driver


akhaku:
Regular schema refreshes are disabled by the driver if it's disabled by the user, see the [Datastax docs](https://docs.datastax.com/en/developer/java-driver/4.13/manual/core/metadata/schema/) for reference. There are a few places where spring-data-cassandra refreshes schema, for example in [SessionFactoryFactoryBean](https://github.com/spring-projects/spring-data-cassandra/blob/4.0.0-M3/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/SessionFactoryFactoryBean.java#L133) and [CqlSessionFactoryBean](https://github.com/spring-projects/spring-data-cassandra/blob/4.0.0-M3/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java#L457). It is probably best to not refresh schema if refreshes are disabled.

Also see #990 for another issue related to schema refreshes.

PR #1255, opened 2022-04-17
Allow disabling schema metadata while performing `SchemaAction`

### LINKED ISSUE #990: Allow disabling schema metadata while performing `SchemaAction` [DATACASS-823]
**[Łukasz Świątek](https://jira.spring.io/secure/ViewProfile.jspa?name=l.swiatek)** opened **[DATACASS-823](https://jira.spring.io/browse/DATACASS-823?redirect=false)** and commented

Cassandra driver 4 changed quite a lot, and now creating tables can take quite a while.

Consider disabling schema metadata on `CqlSession` while performing SchemaAction to reduce the schema creation time

 

```

cqlSession.setSchemaMetadataEnabled(false);

performSchemaAction();

//null to use driver configuration and refresh schema metadata if required

cqlSession.setSchemaMetadataEnabled(null);

```

---
No further details from [DATACASS-823](https://jira.spring.io/browse/DATACASS-823?redirect=false)

akhaku:
Additionally, don't force refresh schemas if schema metadata
is disabled.
Closes #990 and #1253

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
@@ -67,6 +67,7 @@ import com.datastax.oss.driver.api.core.CqlSessionBuilder;
  * @author John Blum
  * @author Mark Paluch
  * @author Tomasz Lelek
+ * @author Ammar Khaku
  * @since 3.0
  */
 public class CqlSessionFactoryBean
@@ -450,11 +451,20 @@ public class CqlSessionFactoryBean
 
 		this.session = buildSession(sessionBuilder);
 
-		executeCql(getStartupScripts().stream(), this.session);
-		performSchemaAction();
+		try {
+			SchemaRefreshUtils.withDisabledSchema(this.session, () -> {
+				executeCql(getStartupScripts().stream(), this.session);
+				performSchemaAction();
+			});
+		} catch (RuntimeException e) {
+			throw e;
+		} catch (Exception e) {
+			throw new IllegalStateException("Unexpected checked exception thrown", e);
+		}
 
-		this.systemSession.refreshSchema();
-		this.session.refreshSchema();
+		if (this.systemSession.isSchemaMetadataEnabled()) {
+			this.systemSession.refreshSchema();
+		}
 	}
 
 	protected CqlSessionBuilder buildBuilder() {
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/SchemaRefreshUtils.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/SchemaRefreshUtils.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.data.cassandra.config;
+
+import com.datastax.oss.driver.api.core.session.Session;
+
+/**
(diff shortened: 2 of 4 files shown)


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
```diff
@@ -450,11 +451,20 @@ public void afterPropertiesSet() {
...
 		this.session = buildSession(sessionBuilder);
 
-		executeCql(getStartupScripts().stream(), this.session);
-		performSchemaAction();
+		try {
+			SchemaRefreshUtils.withDisabledSchema(this.session, () -> {
+				executeCql(getStartupScripts().stream(), this.session);
+				performSchemaAction();
+			});
+		} catch (RuntimeException e) {
```
mp911de, line 459: What's the reason for disabling schema refresh here? And why is there the need to declare and catch exceptions?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
```diff
@@ -450,11 +451,20 @@ public void afterPropertiesSet() {
...
+			throw e;
+		} catch (Exception e) {
+			throw new IllegalStateException("Unexpected checked exception thrown", e);
+		}
 
-		this.systemSession.refreshSchema();
-		this.session.refreshSchema();
+		if (this.systemSession.isSchemaMetadataEnabled()) {
```
mp911de, line 465: Why is the refresh for the actual session removed?

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
```diff
@@ -450,11 +451,20 @@ public void afterPropertiesSet() {
...
+			this.systemSession.refreshSchema();
```
mp911de, line 466: The schema refresh in `SessionFactoryFactoryBean` should be guarded as well.

mp911de (changes requested):
I reviewed the pull request and left a few comments. In general, making schema refreshes conditional on the configuration makes sense.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
akhaku: I believe that's already tackled in this PR, let me know if I missed something!

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
akhaku: It actually happens inside `SchemaRefreshUtils.withDisabledSchema`. It isn't obvious from the PR itself, but from the javadoc of `session.setSchemaMetadataEnabled`:
> If calling this method re-enables the metadata (that is, #isSchemaMetadataEnabled() was false before, and becomes true as a result of the call), a refresh is also triggered.

In other words, if schema refresh becomes re-enabled as a result then a refresh will be triggered.

There is one difference though - previously we'd call `refreshSchema` which is sync and blocking while the schema refresh here is more equivalent to calling `Session#refreshSchemaAsync` and discarding the returned `CompletionStage`. I could update the new code here to wait on the returned future but I don't think it's necessary, let me know if you think otherwise!

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
mp911de: I see. We can revisit exceptions in `performSchemaAction` to remove any unnecessary exceptions. 

> The goal here is to disable schema refresh before applying user-provided schema actions. Datastax4 waits for schema refresh after applying changes and if you're applying multiple it's really slow.

We should leave schema refresh up to the config. If schema refresh is disabled as per config, we don't need to disable it on our side.

mp911de:

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/CqlSessionFactoryBean.java
akhaku: > We should leave schema refresh up to the config
for anyone looking at this thread in the future: my impression based on what you said is that we should rely only on schema being disabled via config in `performSchemaAction`, but I see you added `setSuspendLifecycleSchemaRefresh` and I'm a big fan!

PR #1270, opened 2022-05-19
Allow WriterOptionsBuilder to set ttl to 0.

### LINKED ISSUE #1262: WriteOptions.ttl method no longer accepts zero
The method recently changed from 
`Assert.isTrue(!ttl.isNegative(), "TTL must be greater than equal to zero");
`
to 
`Assert.isTrue(!ttl.isNegative() && !ttl.isZero(), "TTL must be greater than equal to zero");
`

 !ttl.isZero() was added, however, 
1. The message reported still makes it sound like 0 should be allowed
2.  The 0 seems to tell the Cassandra driver to disable TTL:
https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlCreateTable.html#tabProp__cqlTableDefaultTTL

It seems like 0 should be allowed, but if not the error message should be corrected.

tudormarc:
Allow WriterOptionsBuilder to set ttl to 0.

Value 0 tells the Cassandra driver to disable the ttl.

Closes #1262


mp911de:
Thank you for your contribution. That's merged and backported now.

PR #1291, opened 2022-07-19
Add configuration to disable entity lifecycle events

### LINKED ISSUE #1286: Add configuration to disable entity lifecycle events
We should introduce a configuration possibility to disable lifecycle events for entities. Event publishing for large results is costly and in cases where no one is listening, we impose quite some performance drag.

mp911de:
Closes #1286

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/AsyncCassandraTemplate.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/AsyncCassandraTemplate.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
@@ -124,7 +125,7 @@ public class AsyncCassandraTemplate
 
 	private final StatementFactory statementFactory;
 
-	private @Nullable ApplicationEventPublisher eventPublisher;
+	private final EntityLifecycleEventDelegate eventDelegate;
 
 	private @Nullable EntityCallbacks entityCallbacks;
 
@@ -190,11 +191,12 @@ public class AsyncCassandraTemplate
 		this.entityOperations = new EntityOperations(converter);
 		this.exceptionTranslator = asyncCqlTemplate.getExceptionTranslator();
 		this.statementFactory = new StatementFactory(converter);
+		this.eventDelegate = new EntityLifecycleEventDelegate();
 	}
 
 	@Override
 	public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
-		this.eventPublisher = applicationEventPublisher;
+		this.eventDelegate.setPublisher(applicationEventPublisher);
 	}
 
 	@Override
@@ -212,6 +214,18 @@ public class AsyncCassandraTemplate
 	 */
 	public void setEntityCallbacks(@Nullable EntityCallbacks entityCallbacks) {
 		this.entityCallbacks = entityCallbacks;
+	}
+
+	/**
+	 * Configure whether lifecycle events such as {@link AfterLoadEvent}, {@link BeforeSaveEvent}, etc. should be
+	 * published or whether emission should be suppressed. Enabled by default.
+	 *
+	 * @param enabled {@code true} to enable entity lifecycle events; {@code false} to disable entity lifecycle events.
+	 * @since 4.0
+	 * @see CassandraMappingEvent
+	 */
+	public void setEntityLifecycleEventsEnabled(boolean enabled) {
+		this.eventDelegate.setEventsEnabled(enabled);
 	}
 
 	@Override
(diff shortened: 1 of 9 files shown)


christophstrobl (approved):
LGTM

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/AsyncCassandraTemplate.java
```diff
@@ -190,11 +191,12 @@ public AsyncCassandraTemplate(AsyncCqlTemplate asyncCqlTemplate, CassandraConver
 		this.entityOperations = new EntityOperations(converter);
 		this.exceptionTranslator = asyncCqlTemplate.getExceptionTranslator();
 		this.statementFactory = new StatementFactory(converter);
+		this.eventDelegate = new EntityLifecycleEventDelegate();
```
schauder, line 194: Nitpicking: I'd prefer to initialize this on declaration since it does not depend on anything in the constructor.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/AsyncCassandraTemplate.java
```diff
@@ -456,11 +470,12 @@ private ListenableFuture<Boolean> doDelete(Query query, Class<?> entityClass, Cq
 				tableName);
 		SimpleStatement delete = builder.build();
 
-		maybeEmitEvent(new BeforeDeleteEvent<>(delete, entityClass, tableName));
+		maybeEmitEvent(() -> new BeforeDeleteEvent<>(delete, entityClass, tableName));
```
schauder, line 473: I'm actually curious if this does make a difference. After all the closure is an object that needs to be created as well, and I assume a `BeforeDeleteEvent` is a pretty simple object just storing a couple of fields as well.

If it does make a difference I'd love to learn, where the difference comes from.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraTemplate.java
```diff
@@ -183,6 +184,7 @@ public CassandraTemplate(CqlOperations cqlOperations, CassandraConverter convert
 		this.cqlOperations = cqlOperations;
 		this.entityOperations = new EntityOperations(converter);
 		this.statementFactory = new StatementFactory(new QueryMapper(converter), new UpdateMapper(converter));
+		this.eventDelegate = new EntityLifecycleEventDelegate();
```
schauder, line 187: Again: I'd prefer initialization at declaration time when possible.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/ReactiveCassandraTemplate.java
```diff
@@ -190,6 +191,7 @@ public ReactiveCassandraTemplate(ReactiveCqlOperations reactiveCqlOperations, Ca
 		this.cqlOperations = reactiveCqlOperations;
 		this.entityOperations = new EntityOperations(converter);
 		this.statementFactory = new StatementFactory(converter);
+		this.eventDelegate = new EntityLifecycleEventDelegate();
```
schauder, line 194: See above.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/ReactiveCassandraTemplate.java
```diff
@@ -862,9 +876,7 @@ public String getCql() {
 			}
 		}
 
-		return getReactiveCqlOperations()
-				.execute(new GetConfiguredPageSize())
-				.single();
+		return getReactiveCqlOperations().execute(new GetConfiguredPageSize()).single();
```
schauder, line 879: We should protect the previous formatting in this case.

src/main/asciidoc/reference/mapping.adoc
```diff
@@ -530,6 +533,8 @@ Being based on Spring's application context event infrastructure lets other prod
 
 To intercept an object before it goes into the database, you can register a subclass of `org.springframework.data.cassandra.core.mapping.event.AbstractCassandraEventListener` that overrides the `onBeforeSave(…)` method.
 When the event is dispatched, your listener is called and passed the domain object (which is a Java entity).
+Entity lifecycle events can be costly and you may notice a change in the performance profile when loading large result sets.
+You can disable lifecycle events on the Template API object.
```
schauder, line 537: A link to the `setEnabled` methods in the API doc would be nice. At least we should name the method explicitly so the user doesn't have to search for the right method.

schauder (approved):
Looks almost perfect. 

I just have a couple of minor issues that I would change, and one question. 
It's all in the comments.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/AsyncCassandraTemplate.java
mp911de: Probably going onto a very detailed level here. There are two aspects. Allocation-wise, you are probably right. A supplier creates another object. However, I would expect the JIT to inline the supplier code with the `maybeEmit` method and other called methods so that the JIT optimizes the supplier away.

Some events might require further analysis (such as ID extraction) and that would be guarded by the supplier as well to not run the code if the event publishing is disabled.

mp911de:
Merged.

PR #1297, opened 2022-08-20
Refactor userTypeResolver construction into separate method

akhaku:

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/AbstractCassandraConfiguration.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/AbstractCassandraConfiguration.java
@@ -49,6 +49,7 @@ import com.datastax.oss.driver.api.core.CqlSession;
  * @author Matthew T. Adams
  * @author John Blum
  * @author Mark Paluch
+ * @author Ammar Khaku
  */
 @Configuration
 @SuppressWarnings("unused")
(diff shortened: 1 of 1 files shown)


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/AbstractCassandraConfiguration.java
```diff
@@ -260,4 +255,8 @@ protected KeyspacePopulator keyspacePopulator() {
 	protected ByteArrayResource scriptOf(String content) {
 		return new ByteArrayResource(content.getBytes());
 	}
+
+	protected UserTypeResolver userTypeResolver(CqlSession cqlSession) {
+		return new SimpleUserTypeResolver(cqlSession, CqlIdentifier.fromCql(getKeyspaceName()));
```
akhaku, line 260: I should mention that really what I'm trying to do is allow `getKeyspaceName()` to be null, but right now the docs in `AbstractSessionConfiguration` are explicit that it should be non-null... We already override the session factory bean and the only piece that's causing problems with the null keyspace right now is the `UserTypeResolver`, so moving the constructor to a protected method will let me override it without trying to unravel `getKeyspace()`'s nullability

akhaku:
@mp911de I'd love a blessing if you have a moment! Pretty small change, shouldn't have any impact.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/AbstractCassandraConfiguration.java
mp911de: Spring Boot follows the same approach, it would make sense to address this issue as well, see https://github.com/spring-projects/spring-boot/blob/19a7fee1d7d08c081ae1c9f4e4033c8dd71e0bd5/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/cassandra/CassandraDataAutoConfiguration.java#L100.

mp911de:

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/config/AbstractCassandraConfiguration.java
akhaku: Thank you for merging and polishing! I'm not entirely sure what you're getting at, are you just saying that this change is reasonable or that we should explore making a change in spring-boot's CassandraDataAutoConfiguration as well? Or are you saying it would be worth exploring allowing `getKeyspace` to return null?

PR #1322, opened 2022-10-21
Provide smoother usage of Observabilty.

gregturn:
CC @marcingrzejszczak

This one is MUCH easier to use. I also updated the documentation to illustrate its usage.

### DIFF
--- a/src/main/asciidoc/reference/observability.adoc
+++ b/src/main/asciidoc/reference/observability.adoc
@@ -8,3 +8,77 @@ include::{root-target}_conventions.adoc[]
 include::{root-target}_metrics.adoc[]
 
 include::{root-target}_spans.adoc[]
+
+[[observability.registration]]
+== Observability Registration
+
+Spring Data Cassandra is not yet supported in Spring Boot to automatically enable observability.
+Don't worry, we've got you covered!
+Simply add the `@EnableCassandraObservability` annotation to your Spring Boot application, and you'll be all set.
+
+.Activating observability for your Spring Boot application
+====
+[source,java]
+----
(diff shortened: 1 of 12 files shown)


src/main/asciidoc/reference/observability.adoc
```diff
@@ -8,3 +8,80 @@ include::{root-target}_conventions.adoc[]
...
+            repository.save(new Employee("1", "Frodo", "ring bearer"));
+            repository.save(new Employee("2", "Bilbo", "burglar"));
+        };
+    } catch (Exception e) {
+        observation.error(e); // <4>
+        throw e;
+    } finally {
+        observation.stop(); // <5>
+    }
+}
```
ttddyy, line 78: I think wrapping inside of the lambda is more appropriate to capture the observation for this operations.

```java
CommandLineRunner initData(EmployeeRepository repository, ObservationRegistry registry) {  // <1>
    return args -> {
        Observation observation = Observation.start("init-database", registry); // <2>
        try (Observation.Scope scope = observation.openScope()) { // <3>
            repository.save(new Employee("1", "Frodo", "ring bearer"));
            repository.save(new Employee("2", "Bilbo", "burglar"));
        } catch (Exception e) {
            observation.error(e); // <4>
            throw e;
        } finally {
            observation.stop(); // <5>
        }
    };
```

marcingrzejszczak: BTW why not just call `Observation.createNotStarted("init-database"), registry).observe(() -> { repository.save(.....); }` ?

mp911de:

Issue #1335, opened 2022-12-22
Invalid projection definition


gastonlagaf:
Got this on Version: 3.4.5

**Steps For Reproduction:**

Got Entity Class:

```
@Getter
@Setter
@NoArgsConstructor
@AllArgsConstructor
@Table("users_by_email")
public class CassandraUserByEmail extends CassandraEntity<Long> implements User {

    private String email;

    private String phoneNumber;

    private LocationDetails locationDetails;

    @Id
    public String getEmail() {
        return email;
    }
}
```

And repository interface to it:

```
public interface UserByEmailCassandraRepository extends BaseCassandraRepository<CassandraUserByEmail, Long> {

    Optional<User> findByEmail(String value);

}
```

Repository interface parent, supplied with additional interface, required for rest of the repositories
```
public interface BaseCassandraRepository<T extends Entity<ID>, ID extends Serializable> extends CassandraRepository<T, ID>, Repository<T, ID, CriteriaContainer> {
}
```

**Expected Behaviour:**
Mind that CassandraUserByEmail implements User interface. I expect that findByEmail method will return interface implementation, specified in generic.

**Actual Behaviour:**
Error with message
> Failed to instantiate [com.gastonlagaf.springcassandra.entity.User]: Specified class is an interface

**Investigations:**
In CassandraAbstractQuery class there is a snippet (https://github.com/spring-projects/spring-data-cassandra/blob/9c009bc572abf3fda06a28d1a681900cf380305b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/query/AbstractCassandraQuery.java#L105-L111)

Maybe there should be considered a case when returned instance class is interface of domain type

gastonlagaf:
Created PR https://github.com/spring-projects/spring-data-cassandra/pull/1336

mp911de:
> Maybe there should be considered a case when returned instance class is interface of domain type

This is the actual bug. If the domain type is implementing the returned interface, then it is no longer a projection. Instead, we should read the domain type.

PR #1336, opened 2022-12-22
Support domain type interface as return type

gastonlagaf:

Related to issue https://github.com/spring-projects/spring-data-cassandra/issues/1335


mp911de:
Using a slightly different scheme to determine the domain type, see https://github.com/spring-projects/spring-data-cassandra/commit/515bffc784c19f1c67607162043e1f63397b523b.

Issue #1343, opened 2023-01-15
User Defined ConversionService is ignored


neshkeev:
A user can define their own `ConversionService` bean and spring reuses it ([example 1](https://github.com/spring-projects/spring-boot/blob/6567f40650e58aa7da5de437e9c1988a01f53a97/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/ConversionServiceDeducer.java#L50), [example 2](https://github.com/spring-projects/spring-framework/blob/312db368496f317c0cab89684a8a12cc8bb14ad6/spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java#L889)) if the bean name is [`conversionService`](https://github.com/spring-projects/spring-framework/blob/312db368496f317c0cab89684a8a12cc8bb14ad6/spring-context/src/main/java/org/springframework/context/ConfigurableApplicationContext.java#L62) and it's of the `org.springframework.core.convert.ConversionService` type.

Unfortunately `spring-data-cassandra` always creates a new `org.springframework.core.convert.support.DefaultConversionService` in [MappingCassandraConverter#newConversionService](https://github.com/spring-projects/spring-data-cassandra/blob/ad9531d426fb8689e8eceb4d655b911ef9bd53af/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/convert/MappingCassandraConverter.java) which prevents users from tweaking the `conversionService` the way they need.

mp911de:
Let's take a step back before we apply changes and discuss what you want to achieve. The design is by intent encapsulating the conversion service as customizations in the conversion service require additional metadata for proper type-based conversions.

neshkeev:
[...]

PR #1344, opened 2023-01-15
Reuse User Defined Conversion Service.

neshkeev:
Users used to be unable to define their own conversion service beans, because MappingCassandraConverter always used to create a new instance regardless of whether there is a bean with such a type or not. Now an object for conversion service is created only if there are no such a bean defined.

**NOTICE** I don't expect it to be the best solution possible, I'm open for a dialog to make the PR and related code better

Related tickets #1343


mp911de:
Closing as per https://github.com/spring-projects/spring-data-cassandra/issues/1343#issuecomment-1400444829

PR #1388, opened 2023-06-08
Create table using provided name

Mikhail2048:
Hello @mp911de !
As disscussed, I add a new overloaded method with ability to derive the table name that will be created from entity class


mp911de:

PR #1399, opened 2023-06-23
Add exception translation for `DriverTimeoutException`

Mikhail2048:
Cassandra spring-data module does not recogize the `DriverTimeoutException`, which essentially means that driver has timed out during interaction with cluster in general. 
 
Maybe we should consider to not throw `QueryTimeoutException`, but create a separate one, specifically for this case.

@mp911de @schauder please, take a look.


mp911de:
It makes sense to add the mapping. Care to retain the original `cause` and add unit tests?

Mikhail2048:
Of course, I will. So I'll just:
1. Create new Exception mapping for this `DriverTimeoutException` of datastax driver
2. Add a unit test for this case and preserve the cause

Mikhail2048:
@mp911de Done)

mp911de:

Issue #921, opened 2020-04-09
Allow keyspace customization for CassandraPersistentEntity


spring-projects-issues:
**[lmcdasi](https://jira.spring.io/secure/ViewProfile.jspa?name=lmcdasi)** opened **[DATACASS-751](https://jira.spring.io/browse/DATACASS-751?redirect=false)** and commented

In the data stax 4.5 manual it is indicated that creating session/keyspace (https://docs.datastax.com/en/developer/java-driver/4.5/manual/core/) is an anti-pattern:

"`// Anti-pattern: creating two sessions doubles the number of TCP connections opened by the driver`"

You have a solution that uses the AbstractRoutingSessionFactory that just implement that anti-pattern.

But the data stax driver provides actually two different solutions to avoid having that antipattern.
1. use the setKeypace when creating a Statement - but this is OK only if the version supported if > V4. Which would not be available for all.
2. use the built-in QueryBuilder queries that have the keyspace as a parameter. Example: looking at the QueryBuilder.insertInto - it could be changed to use the keyspace since the insertInto signature allows for a a keyspace & a table name. The Statement builder will then build the query by having the keyspace + "." + tablename.

 

The impacted classes are CassandraTemplate & StatementFactory. I attach example code of above where I test the insertTo successfully.

At the same time, while the framework allows me in CassandraConfig to specify a bean to create a cassandraTemplate that can be different than the default CassandraTemplate it does not allow me to re-use the existing framework classes in order to implement my self a template & statement factory where I can specify the keyspace without being forced to copy them out of the framework.

The keyspace can be kept in a ThreadLocal - using the same principle used for AbstractRoutingSessionFactory .

If we cannot have a template as such in the framework, can we at least have the possibility to implement a CassandraTemplace without being forced to copy a lot of code in order to access the existing framework classes ? The template requires quite a few classes that are private.

This assumes that the keyspace schema are the same.

Thank you

---

**Affects:** 3.0 RC1 (Neumann)

**Attachments:**
- [DalAdminTemplate.java](https://jira.spring.io/secure/attachment/27801/DalAdminTemplate.java) (_6.05 kB_)
- [DalStatementFactory.java](https://jira.spring.io/secure/attachment/27800/DalStatementFactory.java) (_39.27 kB_)
- [DalTemplate.java](https://jira.spring.io/secure/attachment/27799/DalTemplate.java) (_39.87 kB_)

1 votes, 4 watchers

mp911de:
[...]

PR #1400, opened 2023-06-26
Implementing keyspace qualification for particular entities 

Mikhail2048:
Hello @mp911de @schauder !
I ahve implemented #921 feature. I have assumed the following behavoir: 

1. There is a possibility to specify a keyspace for an entity in the `@Table` annotation. If it is specified, then for each query we just pass to `QueryBuilder` keyspace specified by user. The goal again is not to mutate _CqlSession_ and avoid creating _CqlSession_ for each keyspace. 
2. On the other hand, if keyspace is omitted or empty, we assume the entity's corresponding table is in the **default for current _CqlSession_** namespace.

I hope it makes sence to you as well. Any suggestions/improvements will be highly appretiated. Thanks!

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/CassandraConnectionFailureException.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/CassandraConnectionFailureException.java
@@ -17,9 +17,14 @@ package org.springframework.data.cassandra;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.springframework.dao.DataAccessResourceFailureException;
+import org.springframework.util.CollectionUtils;
 
 import com.datastax.oss.driver.api.core.metadata.Node;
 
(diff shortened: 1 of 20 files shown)


mp911de:
Have you seen #179? Introducing keyspace to tables raises another set of questions that we tried to address with the previous pull request. For simplicity, let's decouple all polishing within this pull request into a separate one, as the cleanups are almost a no-brainer to merge.

Mikhail2048:
@mp911de No, not yet. But I have checked it now.

1. https://github.com/spring-projects/spring-data-cassandra/pull/179#pullrequestreview-466699019 

Regardign this commet, I think it is the good idea. So we'll just introduce new `getTableInfo()` method that would return a complex object with keyspace and table name. I guess then we need to mark `getTableName()` as `@Deprecated` since accessign table name without keyspace actually loses the sense.

P.S: Also I think we should not expose `setTableName()` to be honest. Ideally, it should be immutable and passed to Cassandra PersistentEntity by the time of it's creation.

2.https://github.com/spring-projects/spring-data-cassandra/pull/179#issuecomment-684514908

I can introduce new overloaded methods to an API of CassandraTemplate to accept this new `TableCoordinates`, no problem.

Any other key components for me to consider?)

By the way, I have separated the polishing commits already. You want me to create a separate PR for this?)

mp911de:
Applying a keyspace to a table sounds straight-forward, but there are a few things to consider:

1. `CassandraTemplate` and the underlying `CqlTemplate` operate within a single keyspace as `CqlSession` is bound to a keyspace. This complicates schema support as we create a table/user type within that keyspace. All of our specifications (e.g. `AlterTableSpecification`, `CreateTableSpecification`) would require another keyspace property to define a different keyspace to generate the appropriate DML statements.
2. Tables using user-defined types from different keyspaces would require an appropriate resolution mechanism. `UserTypeResolver` accepts a name only as it is tied to a single keyspace.

IMO, the keyspace isn't something that is detached from the actual identifier but it rather enriches it. A much more appropriate way to think about this is thinking whether the identifier is qualified or unqualified and having an extension to `CqlIdentifier` in the sense of `QualifiedCqlIdentifier`. Such an approach could allow not to introduce of additional API but rather consider qualified identifiers in e.g. `SimpleUserTypeResolver`, the CQL generator, or our `StatementFactory`. For this to be possible, `CqlIdentifier`'s constructor must be protected to allow subclasses.

mp911de:
I filed https://datastax-oss.atlassian.net/browse/JAVA-3088 asking for constructor rules relaxation.

Mikhail2048:
Yeah, I agree with you. 
I also think we should think of keyspace and table name as a single identifier of the table in cassandra cluster. For this I have created the PR in datastax java driver: https://github.com/datastax/java-driver/pull/1683

mp911de:
Superseded as per https://github.com/spring-projects/spring-data-cassandra/issues/921#issuecomment-2250284092. Thank you for your contribution to the design of customizing keyspaces.

Issue #764, opened 2018-09-12
Cassandra schema validation [DATACASS-594]


spring-projects-issues:
**[tavim](https://jira.spring.io/secure/ViewProfile.jspa?name=tavim)** opened **[DATACASS-594](https://jira.spring.io/browse/DATACASS-594?redirect=false)** and commented

Implement table & column name validation at startup (as SchemaAction enum option, like 'validate' in hibernate)

---
No further details from [DATACASS-594](https://jira.spring.io/browse/DATACASS-594?redirect=false)

PR #1466, opened 2024-01-04
Added schema validation on context startup

mipo256:
This is the PR to solve #764 issue.
The key takeaways in the PR: there is a new class, called `CassandraSchemaValidator`, that is validating Cassandra schema against existing entities in the `MappingContext`. The keyspace where we expect the entities to show up is one that is configured to be used by `CqlSession` (for now that is correct because we do not have a keyspace segregation for entities yet, see this ticket).

The validation checks that:
- All Tables exists
- All **non-transient** properties are expected to be mapped to a single column (I do not know if this is correct or not, appreciate any help on this)
- We expect that the type of the persistent property matches the one we expect in Cassandra  

P.S: I think it would be nice to add this as a bean to Cassandra auto-configuration to spring boot with some properties conditional annotations

### DIFF
--- a/spring-data-cassandra/pom.xml
+++ b/spring-data-cassandra/pom.xml
@@ -233,6 +233,12 @@
 			<scope>test</scope>
 		</dependency>
 
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-test</artifactId>
+			<version>3.1.6</version>
+		</dependency>
+
 	</dependencies>
 
 </project>
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/CassandraKeyspaceDoesNotExistsException.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/CassandraKeyspaceDoesNotExistsException.java
@@ -1,0 +1,15 @@
+package org.springframework.data.cassandra;
+
+import org.springframework.dao.NonTransientDataAccessException;
+
+/**
+ * The exception to be thrown when keyspace that expected to be present is missing in the cluster
+ *
+ * @author Mikhail Polivakha
+ */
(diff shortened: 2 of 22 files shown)


mp911de:
Thanks for having a look. The design requires a bit more thought and not everything should be a `String` or `Exception`. Design-wise, we should decouple validation from any factory beans for a clearer separation of concerns. Later on, once the validation part is fine, we can consider how to integrate validation into application startup.

For type resolution, I suggest you have a look at `DefaultColumnTypeResolver` and `SchemaFactory`, how data types are derived from properties.

Also, validation findings should be part of a model (`ColumnNotFound`, `TableNotFound`) along with references to the underlying CQL item/persistent property/persistent entity.

Let me know whether that helps.

Mikhail2048:
@mp911de I have introduced a couple of changes in accordance to what you've said:
1. `CassandraSchemaValidator` now accepts `CqlSession` directly instead of `CqlSessionFactoryBean`
2. `CassandraSchemaValidator` now also accepts `CassandraConverter` instead of `CassandraMappingContext`. This allwos getting `CassandraMappingContext` from  `CassandraConverter` and also the creation of `DefaultColumnTypeResolver` instance using CassandraConverter`

But I do not fully understand wht you mean here:

> Also, validation findings should be part of a model (ColumnNotFound, TableNotFound) along with references to the underlying CQL item/persistent property/persistent entity.

Can you please give me a glue what is your vision here? You want `CassandraSchemaValidationProfile` to accept not `String` objects as validation error messages but have an abstraction over a message, such as `ColumnNotFound` and `TableNotFound` and so on, am I correct?

Mikhail2048:
I have rebased this PR. Any chance to review it? @mp911de

mp911de:
Currently, this ticket (and several other ones) is on pause because we're limited in bandwidth. We currently can only fix smaller bugs and I don't know when we will have more time for proper enhancements.

Issue #380, opened 2015-04-27
Allow application to start if cassandra is unavailable [DATACASS-210]


spring-projects-issues:
**[Chad Jaros](https://jira.spring.io/secure/ViewProfile.jspa?name=chad.jaros)** opened **[DATACASS-210](https://jira.spring.io/browse/DATACASS-210?redirect=false)** and commented

If cassandra is not available at boot time, CassandraCqlSessionFactoryBean.afterPropertiesSet() causes the application to exit before completing the startup process. It is desirable for my application to complete booting and allow the application to start and report specific errors.

This can be accomplished by making the factory beans operate lazily, retrying the connection on demand and storing the session object once a connection is successfully made

---

**Affects:** 1.2 GA (Fowler)

**Issue Links:**
- [DATACASS-460](https://jira.spring.io/browse/DATACASS-460) NoHostAvailable not caught at bean declaration, but only at initialization
 (_**"is duplicated by"**_)

1 votes, 1 watchers

christophstrobl:
With Spring Boot being the main entry point for configuration of the `SessionFactoryFactoryBean` via the `CassandraDataAutoConfiguration` we should provide means avoid eager initialization of a provided `CqlSession` bean. 
Introducing `Supplier` based alternatives and switching to lazy evaluation would allow the context to start without actually establishing a connection as long as no schema actions need to be performed.

PR #1485, opened 2024-03-15
Allow lenient infrastructure startup. 

christophstrobl:
By avoiding interaction with a potential lazy proxy for a given `CqlSession` until it becomes mandatory we allow users to leverage lazy bean initialization for the `CqlSession` during application startup.

Resolves: #380


mp911de:

Issue #1566, opened 2025-04-08
Add support for Cassandra AOT Repositories


mp911de:
**Supported Features**

* Derived query methods
…

**Limitations**

**Excluded methods**

* `CrudRepository` and other base interface methods
* Querydsl and Query by Example methods
* Methods whose implementation would be overly complex

PR #1574, opened 2025-05-08
Add support for `RepositoryFragmentsContributor`

onobc:
This code proposal adopts the newly added AOT changes ([spring-data-commons/3282](https://github.com/spring-projects/spring-data-commons/pull/3282)) to Cassandra repositories.

> [!NOTE]
> While the AOT repositories support has not been added yet, I think it still makes sense to add these changes in as they will be required by AOT and this is a small enough change to focus on in a single PR.

- Introduces repository fragments contributor and updates the repository factories and factory beans w/ to use the default implementations.

- Also makes the `CassandraRepositoryFactoryBean` consistent with its reactive counterpart by only setting the mapping context from the `CassandraOperations` in `afterPropertiesSet` iff the mapping context was not specified by the user.

See: https://github.com/spring-projects/spring-data-cassandra/issues/1566

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/CassandraRepositoryFragmentsContributor.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/CassandraRepositoryFragmentsContributor.java
@@ -32,7 +32,7 @@ import org.springframework.util.Assert;
  */
 public interface CassandraRepositoryFragmentsContributor extends RepositoryFragmentsContributor {
 
-	CassandraRepositoryFragmentsContributor DEFAULT = DefaultCassandraRepositoryFragmentsContributor.INSTANCE;
+	CassandraRepositoryFragmentsContributor DEFAULT = EmptyFragmentsContributor.INSTANCE;
 
 	/**
 	 * Returns a composed {@code CassandraRepositoryFragmentsContributor} that first applies this contributor to its inputs,
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/CassandraRepositoryFragmentsContributor.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/CassandraRepositoryFragmentsContributor.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.cassandra.repository.support;
+
+import org.springframework.data.cassandra.core.CassandraOperations;
+import org.springframework.data.cassandra.repository.query.CassandraEntityInformation;
+import org.springframework.data.repository.core.RepositoryMetadata;
+import org.springframework.data.repository.core.support.RepositoryComposition.RepositoryFragments;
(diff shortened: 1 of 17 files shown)


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/CassandraRepositoryFragmentsContributor.java
```diff
@@ -0,0 +1,98 @@
...
+	 * @since 5.0
+	 */
+	enum DefaultCassandraRepositoryFragmentsContributor implements CassandraRepositoryFragmentsContributor {
+
+		INSTANCE;
+
+		@Override
+		public RepositoryFragments contribute(RepositoryMetadata metadata,
+				CassandraEntityInformation<?, ?> entityInformation, CassandraOperations operations) {
+			return RepositoryFragments.empty();
```
onobc, line 90: Eventhough the fragment are empty by default for Cassandra, I chose to create a default impl that simple returned empty() to serve as a non-null fragmentsContributor default in the factory/factoryBeans as well as a placeholder in case we add default fragments (e.g. the MongoDB queryDsl fragments).

mp911de: Makes sense to have the same extension points across our Spring Data modules. Having the enum inside the interface makes it public. I suggest having a package-protected visibility for the default implementation.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/ReactiveCassandraRepositoryFragmentsContributor.java
```diff
@@ -0,0 +1,98 @@
...
+	RepositoryFragments contribute(RepositoryMetadata metadata,
+			CassandraEntityInformation<?, ?> entityInformation, ReactiveCassandraOperations operations);
+
+	/**
+	 * Implementation of {@link ReactiveCassandraRepositoryFragmentsContributor} that contributes empty fragments by default.
+	 *
+	 * @author Chris Bono
+	 * @since 5.0
+	 */
+	enum DefaultReactiveCassandraRepositoryFragmentsContributor implements ReactiveCassandraRepositoryFragmentsContributor {
```
mp911de, line 83: Let's move the default impl into its own file to avoid public visibility.

mp911de (changes requested):
Great start, left two comments about visibility.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/support/DefaultCassandraRepositoryFragmentsContributor.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.data.repository.core.RepositoryMetadata;
+import org.springframework.data.repository.core.support.RepositoryComposition;
+
+/**
+ * Implementation of {@link CassandraRepositoryFragmentsContributor} that contributes empty fragments by default.
+ *
+ * @author Chris Bono
+ * @since 5.0
+ */
+enum DefaultCassandraRepositoryFragmentsContributor implements CassandraRepositoryFragmentsContributor {
```
onobc, line 29: Question on naming of "default impls".... what is our team standard on this? I have seen so far quite a bit of `Foo -> FooSupport` rather than `Foo -> DefaultFoo`. I am happy to rename.

mp911de: Mostly, `SimpleFoo` or `DefaultFoo`. In JPA and MongoDB I went for `QuerydslContributor` because the name is already quite long. How about `EmptyContributor` for Cassandra?

onobc: Yeh, I like `EmptyContributor`. I will adjust.

onobc: Done

mp911de:

Issue #1584, opened 2025-06-19
Add support for additional table options (e.g., default_time_to_live) in CassandraAdminTemplate


seungh0:
Hello,

I'd like to propose extending CassandraAdminTemplate to support additional table options, such as default_time_to_live.

For example, the following code currently results in an error:

```kotlin
cassandraAdminTemplate.createTable(false, Table::class.java, mapOf("default_time_to_live" to 180000))

// -> Unable to recognize specified Table option 'default_time_to_live'
```

Additionally, it seems that several valid table options are not currently recognized:

- default_time_to_live
- speculative_retry
- memtable_flush_period_in_ms
- crc_check_chance
- min_index_interval
- max_index_interval
- read_repair
- cdc

These options are documented in the official Cassandra documentation:

- https://cassandra.apache.org/doc/latest/cassandra/reference/cql-commands/create-table.html#table_options
- https://cassandra.apache.org/doc/latest/cassandra/developing/cql/ddl.html#create-table-options
- https://cassandra.apache.org/doc/latest/cassandra/developing/cql/ddl.html#alter-table-statement

Supporting these options in CassandraAdminTemplate would enable developers to define more fine-grained table configurations programmatically, aligning better with native CQL capabilities and increasing flexibility in schema management.

I'm happy to submit a PR if this change aligns with the project's direction.

PR #1586, opened 2025-06-23
Allow use of table options not registered in `TableOption` and extend known `TableOptions`

seungh0:
1. Allow use of table options not registered in TableOption
2. Add support for additional table options

  - default_time_to_live
  - speculative_retry
  - memtable_flush_period_in_ms
  - cdc
  - read_repair
  - min_index_interval
  - max_index_interval
  - crc_check_chance

Closes https://github.com/spring-projects/spring-data-cassandra/issues/1584

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraAdminTemplate.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraAdminTemplate.java
@@ -44,6 +44,7 @@ import com.datastax.oss.driver.api.core.metadata.schema.TableMetadata;
  * @author John Blum
  * @author Vagif Zeynalov
  * @author Mikhail Polivakha
+ * @author Seungho Kang
  */
 public class CassandraAdminTemplate extends CassandraTemplate implements CassandraAdminOperations {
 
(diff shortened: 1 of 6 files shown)


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraAdminTemplate.java
```diff
@@ -127,6 +130,14 @@ public void createTable(boolean ifNotExists, CqlIdentifier tableName, Class<?> e
 		getCqlOperations().execute(CqlGenerator.toCql(createTableSpecification));
 	}
 
+	private void addRawTableOption(String key, Object value, CreateTableSpecification createTableSpecification) {
+		if (value instanceof String) {
+			createTableSpecification.with(key, value, true, true);
+			return;
+		}
+		createTableSpecification.with(key, value, false, false);
+	}
```
seungh0, line 139: In the current situation,  
if users manually apply escaping and quoting,  
it may cause an error in the future if that option is later added to TableOption during a version upgrade.

So I considered two options:

1. Apply escaping and quoting **only when** the option is **not** part of TableOption and the value is a string — as implemented in the PR.  
2. Enhance CqlStringUtils to detect whether the value is already escaped or quoted, and skip escaping if so.

I chose option 1, considering the potential side effects.  
What do you think is the better approach?

mp911de (approved):
Thanks for the decent pull request. I'm going to merge this one for the next release.

spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/CassandraAdminTemplate.java
seungh0: Thank you for your feedback.

PR #1596, opened 2025-08-11
Support multiple `Update` operations for the same column name

### LINKED ISSUE #1525: Update doesn't support multiple set operations for map column
Update stores one column name as key and one operation as value. 

https://github.com/spring-projects/spring-data-cassandra/blob/67d8e597ae900b1c2f1eb145e76755cb25a1bb57/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/query/Update.java#L49

Because of this it is not possible to update several keys for column with map type.

Example of code:
```
CREATE TABLE data (fields map<VARCHAR, VARCHAR>);
```

```
List<Update.AssignmentOp> updateOperations = new LinkedList<>();
updateOperations.add(new Update.SetAtKeyOp(ColumnName.from("map"), "key1", "value1"));
updateOperations.add(new Update.SetAtKeyOp(ColumnName.from("map"), "key2", "value2"));
Update update = Update.of(updateOperations);

OR

Update.empty()
        .set("map").atKey("key1").to("value1")
        .set("map").atKey("key2").to("value2");
```

only one of operations will be in update object.

m3k0813:
Fixes #1525

This PR enhances the `Update` class to support multiple operations on the same column, resolving the limitation of the previous `Map`-based implementation.

* Changed the internal data structure of `Update` from `Map<ColumnName, AssignmentOp>` to `List<AssignmentOp>`. This allows multiple non-conflicting operations (e.g., setting different keys in a map) to coexist.

* Introduced a conflict resolution mechanism to handle duplicate or conflicting operations. The new implementation follows a "last-wins" policy, as suggested in the issue discussion. For example, a `set("name", "B")` operation will replace a previous `set("name", "A")`.

* Added a comprehensive suite of unit tests to `UpdateUnitTests` to verify both the coexistence of non-conflicting operations and the "last-wins" replacement policy for various edge cases.

---

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/query/Update.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/query/Update.java
@@ -17,10 +17,11 @@ package org.springframework.data.cassandra.core.query;
 
 import static org.springframework.data.cassandra.core.query.SerializationUtils.*;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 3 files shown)


spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/query/Update.java
```diff
@@ -215,24 +215,56 @@ public Update decrement(String columnName, Number delta) {
...
+			return e.getIndex() == i.getIndex();
+		}
+
+		return true;
 	}
 
-	@Override
-	public String toString() {
-		return StringUtils.collectionToDelimitedString(updateOperations.values(), ", ");
+	private static boolean equalsNullSafe(@Nullable Object a, @Nullable Object b) {
```
mp911de, line 259: This method isn't needed, use `ObjectUtils.nullSafeEquals(…)` instead.

mp911de (approved):
Thanks a lot. This looks pretty decent. I'm going to polish this pull request for inclusion in the next release.

mp911de:

PR #1652, opened 2026-05-15
Return proxied `Statement` from `ObservationStatement` mutator calls

### LINKED ISSUE #1651: `ObservationRequestTracker` remains open for observed Cassandra statements
## Bug description
When Cassandra request observation is enabled, RequestIdGenerator appears to break the observation wrapper chain.
The problem is that `RequestIdGenerator.getDecoratedStatement(Statement<?> statement, String requestId)` calls:

```java
return statement.setCustomPayload(unmodifiableMap);
```
https://github.com/apache/cassandra-java-driver/blob/c7fe73fe23149c10fe365402b55a6dc88b90b0d4/core/src/main/java/com/datastax/oss/driver/api/core/tracker/RequestIdGenerator.java#L82

Since `Statement<?>` is immutable, `setCustomPayload(...)` returns a **new statement instance**.

If the incoming statement was already wrapped by Spring Data Cassandra observation infrastructure (for example an `ObservationStatement` created from `CqlSessionObservationInterceptor` / tracked by `ObservationRequestTracker` https://github.com/spring-projects/spring-data-cassandra/blob/d41b7ea4691abec94a5b0038f2462c54ad89d6cb/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/observability/CqlSessionObservationInterceptor.java#L169), the returned statement is no longer the wrapper/proxy instance. This drops the observation wrapper and breaks the observation lifecycle, so the corresponding long task timer is never stopped.

In production, this leads to a buildup of open long task timers and memory growth.

## Expected behavior

Observation should be stopped on call to onSucess
https://github.com/spring-projects/spring-data-cassandra/blob/d41b7ea4691abec94a5b0038f2462c54ad89d6cb/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/observability/ObservationRequestTracker.java#L53-L66

## Actual behavior

Applying the request id via `setCustomPayload(...)` creates a new statement instance and loses the observation wrapper. The request still executes, but `ObservationRequestTracker` never closes the corresponding long task timer.

## Confirmed root cause

The exact failure flow is:

1. Spring Data Cassandra wraps the statement for observation.
2. `W3CContextRequestIdGenerator` adds the request id via `statement.setCustomPayload(...)`.
3. Because the statement is immutable, a new plain statement instance is returned.
4. The observation wrapper is lost.
5. Request completes without the original tracked wrapper.
6. `ObservationRequestTracker` does not stop the long task timer.

The relevant method shape is effectively:

```java
default Statement<?> getDecoratedStatement(
    Statement<?> statement, String requestId) {

  Map<String, ByteBuffer> existing = new HashMap<>(statement.getCustomPayload());
  String key = getCustomPayloadKey();
  existing.put(key, ByteBuffer.wrap(requestId.getBytes(StandardCharsets.UTF_8)));

  Map<String, ByteBuffer> unmodifiableMap = Collections.unmodifiableMap(existing);

  return statement.setCustomPayload(unmodifiableMap);
}
```

## Environment

- Spring Data Cassandra: `5.0.5`
- Spring Boot: `4.0.6`
- Micrometer: `1.16.5`
- Java: `25`
- Cassandra driver: `4.19.2`

## Possible fix

https://github.com/spring-projects/spring-data-cassandra/blob/d41b7ea4691abec94a5b0038f2462c54ad89d6cb/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/observability/ObservationStatement.java#L105-L110

```java
Object result = invocation.proceed(); 
if (result instanceof Statement<?> statement) { 
    this.delegate = statement
    return createProxy(this.observation, statement); ;
} 
return result;
```

BenEfrati:

### DIFF
--- a/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/observability/ObservationStatement.java
+++ b/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/observability/ObservationStatement.java
@@ -34,6 +34,7 @@ import com.datastax.oss.driver.api.core.cql.Statement;
  *
  * @author Mark Paluch
  * @author Marcin Grzejszczak
+ * @author Ben Efrati
  * @since 4.0
  */
 final class ObservationStatement implements MethodInterceptor {
@@ -103,9 +104,10 @@ final class ObservationStatement implements MethodInterceptor {
 			}
 		}
 
-		Object result = invocation.proceed();
-		if (result instanceof Statement<?>) {
-			this.delegate = (Statement<?>) result;
+		Object result = invocation.proceed(); 
+		if (result instanceof Statement<?> statement) { 
+			this.delegate = statement;
+			return createProxy(this.observation, statement);
 		}
 
 		return result;
--- a/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/observability/ObservationStatementUnitTests.java
+++ b/spring-data-cassandra/src/test/java/org/springframework/data/cassandra/observability/ObservationStatementUnitTests.java
@@ -15,11 +15,17 @@
  */
 package org.springframework.data.cassandra.observability;
 
-import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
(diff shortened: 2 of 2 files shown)


mp911de:

PR #1665, opened 2026-08-05
Spring Cache using Cassandra

senocak:
A full Spring Cache using Cassandra that implement Spring's cache abstraction on top of Cassandra, storing each entry as a document via the store-independent CassandraTemplate API.

- [ ] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [ ] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [ ] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

### DIFF
--- a/spring-data-cassandra/pom.xml
+++ b/spring-data-cassandra/pom.xml
@@ -86,6 +86,11 @@
 		</dependency>
 
 		<dependency>
+			<groupId>com.fasterxml.jackson.datatype</groupId>
+			<artifactId>jackson-datatype-jsr310</artifactId>
+		</dependency>
+
+		<dependency>
 			<groupId>io.micrometer</groupId>
 			<artifactId>micrometer-test</artifactId>
 			<scope>test</scope>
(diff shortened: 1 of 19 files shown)


mp911de:
What is the rationale? For over a decade, no one asked for a Cassandra-backed cache and the cost of maintenances outweighs benefits, so we'd rather not merge this change.

senocak:
I was using cassandra for one of my clients and using redis or any other distributed cache no needed since cassandra was suitable for it and already have the cassandra running so i wanted to get the all the codes merged into offical codebase. If you think this is not suitable then it can be gladly closed. Thank you.

mp911de:
After iterating over this item, we decided to not merge the pull request. In any case, thanks a lot for your work.

